Issue #39, opened 2023-05-29
Show pod IP for get pod cmd


rougsig:
Looks like [graphql](https://graphql-spec.runpod.io/#definition-PodRuntime) spec can return information about pod ip.

I wanna to create pod with ctl and then connect with ssh to created pod.

pw-git:
I added a PR for this, see above. Please note that depending on factors unknown to me, it might take a while until the IPs and port mapping are available on graphql. So calling `runpodctl get pod` immediately after creating a new pod will sometimes not contain the IPs and ports.
During my tests is seemed to me that for "community" servers it took a bit longer than for the "secure" ones.

PR #50, opened 2023-09-17
feat: IPs and port mappings in pod get output

pw-git:
closes runpod/runpodctl#39


cmd/pod/getPod.go
```diff
@@ -30,6 +30,20 @@ var GetPodCmd = &cobra.Command{
 			}
 			row := []string{p.Id, p.Name, fmt.Sprintf("%d %s", p.GpuCount, p.Machine.GpuDisplayName), p.ImageName, p.DesiredStatus}
 			if AllFields {
+
+				var portEntries int = 0
+				if  p.Runtime != nil && p.Runtime.Ports != nil {
+					portEntries = len(p.Runtime.Ports)
+				}
+				ports := make([]string, portEntries)
+				for j:=0; j<portEntries; j++ {
```
ef0xa, line 39: for j := range ports {

}

cmd/pod/getPod.go
```diff
@@ -38,14 +52,15 @@ var GetPodCmd = &cobra.Command{
 					fmt.Sprintf("%d", p.ContainerDiskInGb),
 					fmt.Sprintf("%d", p.VolumeInGb),
 					fmt.Sprintf("%.3f", p.CostPerHr),
+					fmt.Sprintf("%s", strings.Join(ports[:], "\n") ),
```
ef0xa, line 55: are these supposed to be separated by newlines?

cmd/pod/getPod.go
```diff
@@ -30,6 +30,20 @@ var GetPodCmd = &cobra.Command{
...
+				if  p.Runtime != nil && p.Runtime.Ports != nil {
+					portEntries = len(p.Runtime.Ports)
+				}
+				ports := make([]string, portEntries)
+				for j:=0; j<portEntries; j++ {
+					var privpub string = "prv"
+					if p.Runtime.Ports[j].IsIpPublic {
+						privpub = "pub"
+					}
+					ports[j] = fmt.Sprintf("%s:%d->%d\u00A0(%s,%s)",  p.Runtime.Ports[j].Ip,  p.Runtime.Ports[j].PublicPort,  p.Runtime.Ports[j].PrivatePort,  privpub,  p.Runtime.Ports[j].PortType)
```
ef0xa, line 44: I am deeply suspicious of `0xA0`, the non-breaking space.

DireLines: it's so it works with whatever table writer we're using, we can replace

cmd/pod/getPod.go
DireLines: no

DireLines (approved):
took a while to get to this because we had to implement the same struct for our own purposes, so needed to combine your version and ours. But after checking, it'll work fine. Thanks!

PR #56, opened 2023-12-21
feat: move venv to local disk synced to net vol with tar + some misc updates

DireLines:


DireLines:
additional changes on top of this:

- print more urls
- move network volume selection to start/deploy
- handle migrating old .runpod.yaml to new .runpod/config.toml

PR #66, opened 2024-01-19
[feat] Adds a method to login to the browser

rachfop:
## Demo only

Demos the ability to log in to the browser from the CLI.

Would like to learn more on how certs are generated and how these can be stored in the CLI for auth purposes.

## What does this need

This needs the auth service to actually grab the API key and write in some meaningful way.

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -1,0 +1,25 @@
+package cmd
+
+import (
+	"cli/services"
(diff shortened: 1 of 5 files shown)


Yhlong00:
looks like a good features, should we fix the conflicts and merge it?

PR #83, opened 2024-02-02
Docs review of project cmd

AllysonRosenthal:


cmd/project.go
```diff
@@ -8,8 +8,8 @@ import (
 
 var projectCmd = &cobra.Command{
 	Use:   "project [command]",
-	Short: "manage projects",
-	Long:  "Project management for Runpod projects",
+	Short: "Manage RunPod projects",
+	Long:  "Rapidly develop and deploy a project entirely on RunPod's infrastructure without touching Docker",
```
DireLines, line 12: I like most of these edits, but I feel like this one is kinda sales pitchy

cmd/project/ssh.go
```diff
@@ -41,7 +41,7 @@ func getPodSSHInfo(podId string) (podIp string, podPort int, err error) {
 			return port.Ip, port.PublicPort, nil
 		}
 	}
-	return "", 0, errors.New("no SSH port exposed on pod")
+	return "", 0, errors.New("no ssh port exposed on pod")
```
DireLines, line 44: why only lowercase SSH here?

cmd/project.go
AllysonRosenthal: How about "Develop and deploy a project entirely on RunPod's infrastructure"?

DireLines: That's nice, maybe "Develop and deploy projects" since we used projects plural in the short version

cmd/project/ssh.go
AllysonRosenthal: Good catch. Fixed!

DireLines (approved):

PR #98, opened 2024-02-12
UX Cleanup

justinmerrell:


cmd/project.go
```diff
@@ -8,8 +8,8 @@ import (
 
 var projectCmd = &cobra.Command{
 	Use:   "project [command]",
-	Short: "Manage RunPod projects",
-	Long:  "Develop and deploy projects entirely on RunPod's infrastructure",
+	Short: "(NEW) Manage RunPod projects",
```
DireLines, line 11: rather not say "NEW" here, cause then if the user doesn't update runpodctl it will just always say new and look dumb

DireLines (approved):

PR #100, opened 2024-02-13
feat: create endpoint section in runpod.toml

DireLines:


DireLines:
don't merge yet - need to update endpoint if already exists and config changed, so will need a new gql wrapper for UpdateEndpoint

justinmerrell (approved):

PR #105, opened 2024-02-15
fix: correct verbose for rsync, and dep fix

justinmerrell:


DireLines:
except "building file list ... done" in rsync check and any other lines that are not files

remove jq from required deps

add jq to base image

DireLines (approved):

PR #122, opened 2024-02-27
E-301: Update go mod, add ci, run `gofumpt`

ef0xa:
# [E-301](https://linear.app/runpod/issue/E-301)

I couldn't just add the CI here: go 1.17 changed the format of `go.mod` from `1.16`. We needed to update the version of `go` to get a consistent result, so we might as well hop all the way up to 1.22.

# [E-302](https://linear.app/runpod/issue/E-302) I also unified the module path with go convention, similar to https://github.com/runpod/hapi/pull/15 for HAPI. 

## How I tested it:
[It _is_ the tests.](https://github.com/runpod/runpodctl/actions/runs/8241946315/job/22540107340)


main.go
```diff
@@ -1,6 +1,6 @@
 package main
 
-import "cli/cmd"
```
DireLines, line 3: what's the significance of this change?

DireLines: oh I see you standardized the module name

DireLines: In the package imports local to runpodctl, should the imports be the fully qualified `github.com/...` names or just local paths? I'm unfamiliar with go best practices or any reason to do one or the other

ef0xa: they should always be the fully qualified name. ideally, you should be able to know where to find all the dependencies for a file by just reading the import statement and putting `https://` in front.

DireLines: Alright, that makes sense

DireLines (approved):

PR #137, opened 2024-03-11
Added script for PyTorch issues debugging

kodxana:
Added new command: gpu-test
Curently it's including one of my script to test out PyTorch (PyTorch,go file)
Saves debug informations to /workspace/gpu_diagnostics.json

What script does:
Gather all informations about host including:
- pod id
- machine id
- CUDA version
- PyTorch Version
- GPU model

Runs test on all attached GPU's to makre sure PyTorch fully utilizes CUDA
Logs error messages into .json file for tech support to review.

### DIFF
--- a/cmd/diagnostic/PyTorch.go
+++ b/cmd/diagnostic/PyTorch.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"os"
 	"os/exec"
 	"regexp"
 	"strconv"
@@ -14,6 +13,19 @@ import (
 
 	"github.com/spf13/cobra"
 )
+
+type PodQueryData struct {
+	Query     string            `json:"query"`
+	Variables map[string]string `json:"variables"`
+}
+
+type PodQueryResponse struct {
+	Data struct {
+		Pod struct {
+			MachineID string `json:"machineId"`
+		} `json:"pod"`
+	} `json:"data"`
+}
 
 func getPodMachineID(podID, apiKey string) string {
 	url := fmt.Sprintf("https://api.runpod.io/graphql?api_key=%s", apiKey)
@@ -27,13 +39,21 @@ func getPodMachineID(podID, apiKey string) string {
           }
         }
     `
-	data := map[string]interface{}{
-		"query":     query,
-		"variables": map[string]string{"podId": podID},
+	data := PodQueryData{
+		Query:     query,
+		Variables: map[string]string{"podId": podID},
 	}
-	jsonData, _ := json.Marshal(data)
+	jsonData, err := json.Marshal(data)
+	if err != nil {
+		fmt.Printf("Error marshalling JSON: %v\n", err)
+		return ""
+	}
 
-	req, _ := http.NewRequest("POST", url, bytes.NewBuffer(jsonData))
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(jsonData))
+	if err != nil {
+		fmt.Printf("Error creating request: %v\n", err)
+		return ""
+	}
 	for k, v := range headers {
 		req.Header.Set(k, v)
 	}
@@ -46,12 +66,12 @@ func getPodMachineID(podID, apiKey string) string {
 	}
 	defer resp.Body.Close()
 
-	var result map[string]interface{}
-	json.NewDecoder(resp.Body).Decode(&result)
-	if pod, ok := result["data"].(map[string]interface{})["pod"].(map[string]interface{}); ok {
-		return pod["machineId"].(string)
+	var result PodQueryResponse
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		fmt.Printf("Error decoding response: %v\n", err)
+		return ""
 	}
-	return ""
+	return result.Data.Pod.MachineID
 }
 
 func collectEnvInfo() map[string]string {
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "githubPullRequests.ignoredPullRequestBranches": [
+        "main"
+    ]
+}


cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+		"Content-Type": "application/json",
+	}
+	query := `
+        query Pod($podId: String!) {
+          pod(input: { podId: $podId }) {
+            machineId
+          }
+        }
+    `
+	data := map[string]interface{}{
```
ef0xa, line 30: MINOR: General go HTTP/REST performance tips:
- use a struct here rather than `map[string]any` to avoid extraneous allocations.
- don't make a map to iterate through the headers: just do `req.Header.Set("Content-Type", "application/json")`
- don't make a new `http.Client` for each request: they're safe to share between invocations. if you don't care which client to use, use `http.DefaultClient`
- either or both of the client  or request context should have a timeout.

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+	"os"
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func getPodMachineID(podID, apiKey string) string {
```
ef0xa, line 18: this function should return an error.

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+          }
+        }
+    `
+	data := map[string]interface{}{
+		"query":     query,
+		"variables": map[string]string{"podId": podID},
+	}
+	jsonData, _ := json.Marshal(data)
+
+	req, _ := http.NewRequest("POST", url, bytes.NewBuffer(jsonData))
```
ef0xa, line 36: check this error and return it.

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+func getPodMachineID(podID, apiKey string) string {
+	url := fmt.Sprintf("https://api.runpod.io/graphql?api_key=%s", apiKey)
+	headers := map[string]string{
+		"Content-Type": "application/json",
+	}
+	query := `
```
ef0xa, line 23: OPTIONAL: this can be a `const`.

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		fmt.Printf("Failed to fetch machineId: %v\n", err)
+		return ""
+	}
+	defer resp.Body.Close()
+
+	var result map[string]interface{}
+	json.NewDecoder(resp.Body).Decode(&result)
+	if pod, ok := result["data"].(map[string]interface{})["pod"].(map[string]interface{}); ok {
```
ef0xa, line 51: again, don't use `map[string]any`: just make a struct type that's the shape you want. in this case, it would be 

```go
type Result struct { 
    Data struct  { 
        Pod struct { 
            MachineID string  `json:"machineId"` 
         } `json:"pod"`
    } `json:"data"`
 }
```

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+		return pod["machineId"].(string)
+	}
+	return ""
+}
+
+func collectEnvInfo() map[string]string {
```
ef0xa, line 57: (MINOR, OPTIONAL): again, go style here is to use struct types rather than maps, since they involve no heap allocations.

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+func parseNvidiaSMIOutput(output string) map[string]string {
+	cudaVersionRegex := regexp.MustCompile(`CUDA Version: (\d+\.\d+)`)
+	driverVersionRegex := regexp.MustCompile(`Driver Version: (\d+\.\d+\.\d+)`)
+	gpuNameRegex := regexp.MustCompile(`\|\s+\d+\s+([^\|]+?)\s+On\s+\|`)
+
+	cudaVersion := cudaVersionRegex.FindStringSubmatch(output)
+	driverVersion := driverVersionRegex.FindStringSubmatch(output)
+	gpuName := gpuNameRegex.FindStringSubmatch(output)
+
+	info := map[string]string{
```
ef0xa, line 85: (MINOR, OPTIONAL): see above re: structs vs maps

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+	jsonData, _ := json.Marshal(data)
+
+	req, _ := http.NewRequest("POST", url, bytes.NewBuffer(jsonData))
+	for k, v := range headers {
+		req.Header.Set(k, v)
+	}
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
```
ef0xa, line 43: use `return fmt.Errorf` with `%w` instead of just printing the error.

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+func getNvidiaSMIInfo() map[string]string {
+	cmd := exec.Command("nvidia-smi")
+	output, err := cmd.Output()
+	if err != nil {
+		return map[string]string{"Error": fmt.Sprintf("Failed to fetch nvidia-smi info: %v", err)}
+	}
+	return parseNvidiaSMIOutput(string(output))
+}
+
+func getSystemInfo() map[string]interface{} {
```
ef0xa, line 113: (OPTIONAL, MINOR): this function is tiny and only called once. inline it.

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+			results[fmt.Sprintf("GPU %d", gpuID)] = fmt.Sprintf("Error: %v", err)
+		} else {
+			results[fmt.Sprintf("GPU %d", gpuID)] = strings.TrimSpace(string(output))
+		}
+	}
+
+	return results
+}
+
+func saveInfoToFile(info map[string]interface{}, filename string) {
```
ef0xa, line 154: (OPTIONAL, MINOR): this function is tiny and only called once. inline it.

cmd/diagnostic/PyTorch.go
```diff
@@ -0,0 +1,171 @@
...
+		info["Driver Version"] = driverVersion[1]
+	}
+	if len(gpuName) > 1 {
+		info["GPU Name"] = strings.TrimSpace(gpuName[1])
+	}
+
+	return info
+}
+
+func getNvidiaSMIInfo() map[string]string {
```
ef0xa, line 104: (OPTIONAL, MINOR): this function is tiny and only called once. inline it.

ef0xa (changes requested):
This mostly looks very good. Have a lot of small comments on general performance and 'go/house style' that are just meant to be pointers. Feel free to ignore the stuff labeled OPTIONAL or MINOR: just keep them in mind for next time.

DireLines:
What will happen if people try to run this on the client? Is there anything telling them that it is a diagnostic to be run specifically through web terminal on the pod? I think somebody will try to run it locally and be confused by the output if not

kodxana:
> What will happen if people try to run this on the client? Is there anything telling them that it is a diagnostic to be run specifically through web terminal on the pod? I think somebody will try to run it locally and be confused by the output if not

Most of the output in json file would be empty though script will try to run PyTorch test.

PR #141, opened 2024-04-12
release artifact improvements

ef0xa:
- publish binaries for android and linux
- strip and compress released binaries using `upx`
- use embedding rather than `-ldflags` to put version info in

## How I tested it
I can't, except by trying a release.


DireLines (approved):
lgtm need to test by running release action

PR #151, opened 2024-06-24
E-951: new command `scp-help`: teaches people how to use scp

ef0xa:
# E-951 - teach people how to use SCP

Add new command, `scp-help`, which teaches users how to use SCP. Since it will show up in the command list, that should make this behavior more discoverable.

I also slightly cleaned up `send` and `receive` so that they don't fail with index-out-of-range errors and added an (extremely generous) timeout for for looking up the relays so it doesn't just hang forever.

## How I tested it
Nothing to test.

### DIFF
--- a/cmd/croc/send.go
+++ b/cmd/croc/send.go
@@ -1,11 +1,14 @@
 package croc
 
 import (
+	"errors"
 	"fmt"
+	"log"
 	"math/rand"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/schollz/croc/v9/src/models"
 	"github.com/schollz/croc/v9/src/utils"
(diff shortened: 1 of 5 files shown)


cmd/croc/send.go
```diff
@@ -27,20 +30,27 @@ var code string
...
 var SendCmd = &cobra.Command{
-	Use:   "send [filename(s) or folder]",
-	Args:  cobra.ExactArgs(1),
+	Use:   "send [file0] [file1] ...",
+	Args:  cobra.MinimumNArgs(1),
 	Short: "send file(s), or folder",
 	Long:  "send file(s), or folder to pod or any computer",
-	Run: func(cmd *cobra.Command, args []string) {
-
-		rand.Seed(time.Now().UnixNano())
```
ef0xa, line 36: no longer necessary (and in fact _reduces_ the randomness rather than enhancing it in `go >= 1.22` : https://go.dev/blog/chacha8rand

DireLines (approved):

Issue #156, opened 2024-07-01
can't start spot pod using the cli


youcefs21:
I'm trying to start a pod using the cli by doing:
```
 runpodctl start pod <id>
```
but I'm getting the error: `Error: Cannot resume a spot pod as an on demand pod.`

I also tried putting a bid that matches the spot pricing:
```
runpodctl start pod --bid 0.340 <id>
```
and that gets me a different error `Error: PodBidResume: statuscode 400`

youcefs21:
turns out the graphQL call was missing an argument, fixed it in #157

PR #157, opened 2024-07-01
feat and fix: add gpuCount argument to StartSpotPod

youcefs21:
# Add gpuCount argument

## Related:
- #156: can't start spot pod using the cli 

## Description

The start StartSpotPod endpoint is missing the gpuCount argument, causing it to not work at all. I added the argument to the ctl, and the corresponding graphQL call.

## How I tested it

I run the graphQL query by hand with and without the argument I added. Without it, I get back a 400 error, and with it everything works great.

### DIFF
--- a/api/pod.go
+++ b/api/pod.go
@@ -355,11 +355,11 @@ func StartOnDemandPod(id string) (pod map[string]interface{}, err error) {
 	return
 }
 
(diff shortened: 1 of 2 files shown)


Yhlong00 (approved):

voccer-pionero:
This PR has not been officially released yet?

PR #159, opened 2024-09-08
E-1654: fix: only add key when not exist in db.

Yhlong00:
# E-1654: Only Add Key If It Doesn’t Already Exist in the Database

# Description
Currently, running `runpodctl config --apiKey xxx` creates a new key and appends it to the `pubKey` field each time.

Change: Modify the behavior so that a key is only created if it doesn’t already exist in the database.

# How do I test it?
Build package locally and run command `config --apiKey xxx`

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -2,7 +2,7 @@ package config
 
 import (
 	"fmt"
-	"os"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


DireLines (approved):
tested all the new paths. works in all cases

Issue #174, opened 2024-11-01
The Linux installation snippet yields the error "Unknown option 'show-progress'"


MasonMcGill:
Following the Linux instructions (`wget -qO- cli.runpod.net | sudo bash`) results in an error on Fedora 41:

```shell
$ wget -qO- cli.runpod.net | sudo bash
Installing runpodctl...
All system requirements satisfied.
Latest version of runpodctl: v1.14.4
Unknown option 'show-progress'
```

jasper-zanjani:
I also got this on Fedora 40 which has wget 2.2.0. On line 122 the call to wget should be changed to

```sh
wget --progress=bar
```

PR #177, opened 2025-01-16
Fixed call to wget

jasper-zanjani:
This is a minor fix for the call to wget in the installation script (issue [174](https://github.com/runpod/runpodctl/issues/174))


DireLines (approved):
thanks!

Issue #11, opened 2022-10-28
Unable to use podRentInterruptable


nbanks:
Using [the web interface](https://www.runpod.io/console/gpu-cloud) I'm able to Deploy a spot Instance instead of an On Demand instance.  It would be nice to be able to do this using the command line tool too.

I tried naively replacing podFindAndDeployOnDemand with podRentInterruptable, but this failed.  I have no idea if this was a permission problem, a server problem, or a client problem.  (If I could get it to work, I'd provide a pull request.)  I can see the current spot price using `runpodctl get cloud`.  Once I create a pod through the web, I am able to see it and stop it using the command line interface.  
[I found this documentation.](https://graphql-spec.runpod.io/#mutation-podRentInterruptable)

zhl146:
You can do this via the gql api: https://docs.runpod.io/graphql-api/create-pod#create-spot-pod

This was never added to the cli, though. Feel free to add it via PR an we'll add it eventually!

PR #186, opened 2025-03-22
feat: Create spot pod by adding bidPerGpu argument

raghuveer-s:
# Create Spot Pod

The create command can only create on-demand pods at the moment since it uses `podFindAndDeployOnDemand` in the GraphQL API underneath. If we want to create spot pods, we need to use the `podRentInterruptable` mutation which needs the `bidPerGpu` field as part of its input.

## How I tested it
Manually tested and verified with the RunPod dashboard as well. 
There are a few cases to watch for:
* If bidPerGpu flag does not exist, then use the on-demand pod creation flow.
* If bidPerGpu flag exists and it is <= 0, then return an error.
* If bidPerGpu flag exists and it is > 0, then attempt to create a spot pod using the GraphQL API.

### DIFF
--- a/api/pod.go
+++ b/api/pod.go
@@ -155,10 +155,34 @@ type CreatePodInput struct {
 	Ports             string    `json:"ports"`
 	SupportPublicIp   bool      `json:"supportPublicIp"`
 	StartSSH          bool      `json:"startSsh"`
+	Spot              bool      `json:"spot"`
 	TemplateId        string    `json:"templateId"`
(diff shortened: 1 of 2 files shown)


raghuveer-s:
Related issue: https://github.com/runpod/runpodctl/issues/11

justinwlin:
Thanks! Sorry for getting back after so long, Im currently in the process to own this repository and help get things moved in. Ill be looking deeper into this coming week or so, to see what I can do to get this merged in.

Issue #190, opened 2025-06-01
Global networking option


YayL:
Why is there no global networking option?

TimPietruskyRunPod:
Resolved in [v2.0.0](https://github.com/runpod/runpodctl/releases/tag/v2.0.0).

```bash
runpodctl pod create \
  --template-id <template-id> \
  --gpu-id "NVIDIA A100 80GB PCIe" \
  --global-networking \
  --cloud-type SECURE \
  --volume-in-gb 50 \
  --name my-pod
```

The `--global-networking` flag enables global networking on pod creation. Requires secure cloud and a GPU pod.

PR #191, opened 2025-06-01
Added globalNetwork as a flag

YayL:
# Added globalNetwork as a flag

This patch adds the `globalNetwork` flag to allow global networking to be enable (when applicable) throught the CLI tool. This was not previously possible and I did not understand why so I just quickly fixed it and thought that someone else might also need this.

(Fix for #190)

## How I tested it
I tested it by creating two pods using the following commands:
```
go run main.go create pod --name "testing" --imageName "runpod/base:0.5.1-cpu" --gpuType "NVIDIA RTX A4000" --startSSH --secureCloud --globalNetwork
go run main.go create pod --name "testing2" --imageName "runpod/base:0.5.1-cpu" --gpuType "NVIDIA RTX A4000" --startSSH --secureCloud --globalNetwork
```
Then testing if the two pods could ping each other as per the global networking [documentation](https://docs.runpod.io/pods/networking)

### DIFF
--- a/api/pod.go
+++ b/api/pod.go
@@ -139,6 +139,7 @@ func GetPods() (pods []*Pod, err error) {
 }
 
 type CreatePodInput struct {
+	GlobalNetwork  bool 		`json:"globalNetwork"`
 	CloudType         string    `json:"cloudType"`
 	ContainerDiskInGb int       `json:"containerDiskInGb"`
 	DeployCost        float32   `json:"deployCost,omitempty"`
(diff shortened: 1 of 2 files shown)


justinwlin:
Thanks! Sorry for getting back after so long, will be looking to move this through the team soon :) I like this change.

justinwlin:
Unable to verify this works personally, and I think there are some more nuances need to look into. Ive created a ticket for us to use internally and this as a reference. Thanks!

PR #200, opened 2025-09-17
FIX: invalid header field value for "User-Agent". Process version string to strip possible line break.

demetrio108:
After installing latest version of runpodctl, I was getting error while using API:

```
net/http: invalid header field value for "User-Agent"
```

After debugging I found out that the error is due to line break in 'version' file. Simple solution was to remove line break, but I think it's better to remove line breaks programmatically, as this line breaks are very easy to introduce.

### DIFF
--- a/api/query.go
+++ b/api/query.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"runtime"
+	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


DeJayDev (approved):
Thank you for your contribution!

PR #201, opened 2025-09-17
Fix version string and auth header

brosenpod:
# Fix version string and auth header

If the version string ends with a newline it will cause this error: `net/http: invalid header field value for "User-Agent"`
Strip trailing newlines from the version.

Sending the API key via a parameter is deprecated. Send it in an auth header instead.

## How I tested it

Local dev

### DIFF
--- a/api/query.go
+++ b/api/query.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"runtime"
+	"strings"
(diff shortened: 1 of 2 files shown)


deanq (approved):
Thanks! I missed this when I moved our codebases to use Auth Bearer token headers.

PR #210, opened 2025-11-20
CE-989 and CE-990

brennentsmith-runpod:
CE-989:
Right now, we support a custom "secret code" via send (--code) However this results in a literal insertion which overrides our custom code definition. 

This results in receive failing as we expect 4 hyphens in order to work. The only hyphens that matter is the final one which denotes the relay. 

Correct this behavior so we accept custom codes. 

CE-990:
Right now we use a hardcoded version file which makes version incrementing difficult as one has to remember to manually increment it. 

Move to a programatic version where we use the following form:
`<version number>-<git hash>`

This way we can verify what the codebase was at the time of build.

### DIFF
--- a/cmd/croc/receive.go
+++ b/cmd/croc/receive.go
@@ -59,7 +59,6 @@ var ReceiveCmd = &cobra.Command{
 
 		if relayIndex < 0 || relayIndex >= len(relays) {
 			log.Fatalf("Relay index %d not found; please retry 'runpodctl send' to generate a valid code.", relayIndex)
-			return
 		}
 		relay := relays[relayIndex]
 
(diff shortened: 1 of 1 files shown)


Yhlong00 (approved):

cmd/croc/receive.go
```diff
@@ -48,22 +48,26 @@ var ReceiveCmd = &cobra.Command{
...
+			return
+		}
 		relay := relays[relayIndex]
 
 		crocOptions := croc.Options{
 			Curve:         "p256",
 			Debug:         false,
 			IsSender:      false,
-			NoPrompt:      true,
+			NoPrompt:      true,
```
brosenpod: Nit: Trailing whitespace

brennentsmith-runpod: Good catch - fixing

PR #214, opened 2025-11-22
Fix: runpodctl send/receive "room (secure channel) not ready" error

acamq:
Using runpodctl send/recieve resulted in this error:
```bash
securing channel...runpodctl-receive: croc: receive: room (secure channel) not ready, maybe peer disconnected
```
But using croc directly worked fine. 

The sender and receiver were using fundamentally different croc implementations:

* Sender (send.go): Used the local forked croc implementation from cmd/croc/croc.go
* Receiver (receive.go): Used the external library github.com/schollz/croc/v9/src/croc

The receiver now uses the local fork of the croc implementation. Additionally, missing configuration options were added to the receiver side. 

## How I tested it

Built on Zorin OS 18 with Linux kernel 6.14.0-33-generic, x86_64. Tested by sending and receiving files to and from a pod.


brennentsmith-runpod:
Thank you for the report! Your assessment was correct and this has been fixed in release 1.14.15-dac76ad - this will be rolling out to the production environment this week. 

Fixed release: https://github.com/runpod/runpodctl/releases/tag/v1.14.15
Change: https://github.com/runpod/runpodctl/pull/217

PR #219, opened 2025-12-29
Fix templateId not applying disk and volume settings from template

### LINKED ISSUE #163: Specifying --templateId does not apply disk and volume size/path settings
The container disk / volume mount path / volume size are not applied even though they are already defined in the template.

pantafive:
## Summary
- When `--templateId` is specified, disk/volume settings from template are now used instead of being overwritten by CLI flag defaults

## Problem
CLI flags have non-zero defaults (`--containerDiskSize=20`, `--volumeSize=1`, `--volumePath=/runpod`) that were always sent to the API, overriding template settings even when user didn't explicitly set them.

## Solution
- Added `omitempty` to `containerDiskInGb`, `volumeInGb`, `volumeMountPath` JSON tags
- Added `TemplateFieldFlags` struct and `ClearUnchangedTemplateFields` method to clear these fields when templateId is set and flags were not explicitly changed
- Zero values are now omitted from JSON, allowing API to use template values

Fixes #163

### DIFF
--- a/api/pod.go
+++ b/api/pod.go
@@ -140,7 +140,7 @@ func GetPods() (pods []*Pod, err error) {
 
 type CreatePodInput struct {
 	CloudType         string    `json:"cloudType"`
-	ContainerDiskInGb int       `json:"containerDiskInGb"`
+	ContainerDiskInGb int       `json:"containerDiskInGb,omitempty"`
 	DeployCost        float32   `json:"deployCost,omitempty"`
 	DockerArgs        string    `json:"dockerArgs"`
 	DataCenterId      string    `json:"dataCenterId"`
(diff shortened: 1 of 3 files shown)


jojje (commented):
LGTM. 

Can't think of a better way to balance the convenience of having "sensible defaults" in the CLI, _and_ for being able to avoid having to _explicitly_ repeat whatever was already defined in a template. So this seems a good compromise to me since **if** a user specifies a template, then it's most likely they **intend** for whatever is configured in the template to be used _rather_ than whatever happens to be the default for the CLI options.

Edit: And the _very issue_ this PR is addressing has actually bitten me personally on more than one occasion, so I'd be glad to see it merged.

PR #230, opened 2026-02-06
fix: Remove unnecessary hash for Model Repo uploads and bump Go version

brosenpod:
# Remove unnecessary hash for Model Repo uploads and bump Go version

Model Repo uploads do not accept a hash, causing an error if they are specified. Do not specify them.

Bump Go version as well.

## How I tested it

Local dev

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/runpod/runpodctl
 
-go 1.24
+go 1.25.7
 
 require (
 	github.com/denisbrodbeck/machineid v1.0.1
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/runpod/runpodctl
 
-go 1.24
+go 1.25.7
```
galen-runpod: please bump the github workflow runner(s) as well.

galen-runpod (approved):

PR #232, opened 2026-02-24
feat(pod): add list filters and default to running pods

TimPietruskyRunPod:
## Summary
- Add `--status`, `--since`, `--created-after`, and `--all` flags to `pod list`
- Default behavior now shows only running pods (like `docker ps`)
- Use `--all` or `--status exited` to see non-running pods
- Slim down output fields to essential info (id, name, status, image, gpu, cost)

## Usage
```bash
runpodctl pod list                              # running pods only (default)
runpodctl pod list --all                        # all pods including exited
runpodctl pod list --status exited              # filter by status
runpodctl pod list --since 24h                  # pods created in last 24 hours
runpodctl pod list --created-after 2025-01-15   # pods created after date
```

## Test plan
- [x] Unit tests pass (`go test ./cmd/pod/...`)
- [ ] E2E: `pod list` returns only running pods
- [ ] E2E: `pod list --all` includes exited pods

### DIFF
--- a/cmd/pod/list.go
+++ b/cmd/pod/list.go
@@ -1,6 +1,11 @@
 package pod
 
 import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
 	"github.com/runpod/runpodctl/internal/api"
 	"github.com/runpod/runpodctl/internal/output"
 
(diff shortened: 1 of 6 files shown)


TimPietruskyRunPod:
Addressed all review feedback:

- **Dead API fields (High):** Removed `IncludeMachine`/`IncludeNetworkVolume` from `PodListOptions` and the API layer
- **parseDuration (Medium):** Rewrote to support stdlib formats (`30m`, `1h30m`) with `d` as the only custom suffix, plus positive-value validation
- **--all + --status conflict (Medium):** Returns error when both flags are used together
- **Missing tests (Medium):** Added table-driven unit tests for `parseDuration` and `parseCreatedAt`
- **CreatedAt type (Low):** Normalized from `interface{}` to RFC3339 `string`
- **Redundant ToUpper (Low):** Cleaned up since `EqualFold` handles case already

@max4c ready for another look

max4c (approved):
addressed comments and review lgtm

PR #233, opened 2026-02-24
chore(docs): update agents.md and symlink claude.md

TimPietruskyRunPod:
## Summary
- Update AGENTS.md with improved build instructions (`go install .` + copy to `/usr/local/bin`)
- Add e2e test command guidance (`go test -tags e2e ./...`)
- Add note about keeping the runpodctl skill in sync when commands change
- Replace CLAUDE.md with a symlink to AGENTS.md — single source of truth, zero drift

## Test plan
- [x] Symlink resolves correctly (`cat CLAUDE.md` shows AGENTS.md content)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,12 +76,7 @@ runpod/
 ## build commands
 
 ```bash
-# install globally (always do this after changes)
-go install .
-# also copy to /usr/local/bin so both paths are up to date:
-sudo cp ~/go/bin/runpodctl /usr/local/bin/runpodctl
-
-# local development build
+# local development build (always do this after changes)
 make local
 # output: bin/runpod
 
@@ -92,7 +87,7 @@ make release
 # run unit tests
 go test ./...
 
-# run all tests (unit + e2e) — always run after changes
+# run e2e tests (requires RunPod API credentials)
 go test -tags e2e ./...
 ```
 


max4c:
## PR Review

Overall this is a clean, focused docs/config PR. The intent is solid — consolidating build guidance and creating a single source of truth between AGENTS.md and CLAUDE.md. A few things worth addressing before merging:

---

### Issues

**1. `go install .` is not equivalent to `make local` — it should not be the primary recommended build step**

The PR positions `go install .` as the default build step ("always do this after changes"), but this silently omits the `-ldflags` version injection that `make local` and `make release` use. Running `go install .` produces a binary where `runpodctl version` will show an empty or incorrect version string. For day-to-day development, `make local` is the correct primary command. The `go install .` step can be mentioned as an alternative for getting a binary on `$PATH`, but should not be presented as the canonical build step.

**2. The `sudo cp` instruction is unnecessary and fragile**

`go install .` already places the binary at `$GOPATH/bin/runpodctl` (default: `~/go/bin/runpodctl`). If `~/go/bin` is on `$PATH` (which the Go installer sets up by default), there is no reason to `sudo cp` into `/usr/local/bin`. This step also assumes a specific `GOPATH` location (`~/go/bin`), which is not guaranteed. Recommend removing this line or, at minimum, making it conditional with a note like "only if `~/go/bin` is not on your PATH".

**3. "always run after changes" for `go test -tags e2e ./...` is misleading**

E2E tests require live RunPod API credentials and running infrastructure. Blanket guidance to "always run after changes" will break CI for contributors without those credentials and is not standard practice for a local dev workflow. This should be qualified — e.g., "run before submitting a PR if you have RunPod API access configured".

---

### Suggestions

**4. CLAUDE.md symlink — potential tooling compatibility issue**

Converting CLAUDE.md from a redirect file to a symlink works in most environments, but some tools (certain CI runners, Windows Git clients, GitHub's own raw file viewer) do not follow symlinks in repositories. The previous approach of a one-line redirect file was already a working single-source-of-truth pattern. If a symlink is used, it is worth noting this in the PR and confirming it resolves correctly in the target CI environment. At minimum, the test plan should include verifying behavior in the CI pipeline, not just locally.

**5. Skill sync note belongs in a "Contributing" section, not "important notes"**

The note about keeping the runpodctl skill at `https://github.com/runpod/skills` in sync is useful, but it is an internal/contributor concern. Placing it first in the "important notes" section (above the never-start-servers note) gives it higher prominence than operational runtime notes. Consider moving it to a dedicated "Contributing" or "Maintenance" section.

---

### Summary

| # | Severity | Item |
|---|----------|------|
| 1 | Medium | `go install .` missing version ldflags — misleading as primary build step |
| 2 | Low | `sudo cp ~/go/bin/runpodctl /usr/local/bin` is fragile and unnecessary |
| 3 | Medium | E2E test "always run" guidance is too broad for contributors without credentials |
| 4 | Low | Symlink compatibility in CI/Windows environments unverified |
| 5 | Nit | Skill sync note placement in "important notes" |

Items 1 and 3 are the most important to address before merging, as they could lead contributors to produce misconfigured builds or broken CI runs.

max4c:
## Re-review Follow-up

Thanks for the updates, @TimPietruskyRunPod. Here is a status breakdown of every item from the original review.

---

### Resolved

**Issue 1 (Medium) — `go install .` missing version ldflags**
Fixed. The `go install .` block has been removed. `make local` is now correctly presented as the primary build step, and the comment "(always do this after changes)" is appropriately placed on that command.

**Issue 2 (Low) — `sudo cp ~/go/bin/runpodctl /usr/local/bin` fragile and unnecessary**
Fixed. The entire `sudo cp` line is gone along with the `go install .` block.

**Issue 3 (Medium) — E2E test "always run" guidance too broad**
Fixed. The comment now reads `# run e2e tests (requires RunPod API credentials)`, which correctly sets expectations for contributors without live credentials.

---

### Still Outstanding

**Issue 4 (Low) — Symlink compatibility in CI/Windows environments**
Not addressed. CLAUDE.md is still converted to a symlink (`mode 120000`). This works on Linux/macOS but can break on Windows Git clients (depending on `core.symlinks` config) and may render unexpectedly in some CI environments or GitHub's raw file viewer. The previous one-line redirect file (`See AGENTS.md`) was already a working single-source-of-truth approach without the compatibility tradeoff. This is low severity, but worth a conscious decision rather than leaving it unverified.

**Issue 5 (Nit) — Skill sync note placement**
Not addressed. The `keep the skill in sync` note remains as the first bullet in the "important notes" section, above the operational runtime notes (`never start/stop servers`, etc.). This is a contributor/maintenance concern, not a runtime operational note. Moving it to a dedicated "Contributing" or "Maintenance" section would keep the priority ordering of notes clear for agents reading the file. This is a nit — not a blocker.

---

### New Observation

The `make local` comment now reads "always do this after changes", which is accurate for local dev. However, the `## build commands` section no longer explains what binary is actually placed on `$PATH` for interactive use after `make local` (the output is `bin/runpod`, not `runpodctl`). A contributor following this file alone would need to know to either add `bin/` to their PATH, rename/alias the binary, or use a separate mechanism to test the binary as `runpodctl`. This may be obvious to existing contributors but is worth a one-line note for new contributors.

---

### Summary

| # | Severity | Item | Status |
|---|----------|------|--------|
| 1 | Medium | `go install .` missing ldflags | Resolved |
| 2 | Low | `sudo cp` fragile/unnecessary | Resolved |
| 3 | Medium | E2E "always run" too broad | Resolved |
| 4 | Low | Symlink compatibility unverified | Still outstanding |
| 5 | Nit | Skill sync note placement | Still outstanding |
| 6 | Nit | No guidance on using `bin/runpod` binary post-build | New / minor |

The two medium issues are resolved. Items 4–6 are all low/nit severity. This is close to mergeable — item 4 is the only one I would recommend explicitly deciding on (symlink vs. redirect file) rather than leaving as implicit.

PR #260, opened 2026-03-23
fix: use correct gpu id in pod create help examples

TimPietruskyRunPod:
## Summary

- **GPU ID fix**: help examples used `"NVIDIA RTX 4090"` which is not a valid GPU ID and causes pod creation to fail. Corrected to `"NVIDIA GeForce RTX 4090"` (from `runpodctl gpu list`)
- **Docker image fix**: replaced outdated `runpod/pytorch:2.1.0-py3.10-cuda11.8.0-devel-ubuntu22.04` with `runpod/pytorch:1.0.3-cu1281-torch291-ubuntu2404`
- **Regenerated docs**: `docs/` markdown files regenerated (also picks up missing `--network-volume-id` flag from PR #259)

### Also needs fixing (separate repos)

- **Skill** (`runpod/skills/runpodctl/SKILL.md`): same GPU ID + image + completion instructions — fix ready locally, needs its own PR
- **Docs PR** (runpod/docs#563): uses the same wrong GPU ID and image across multiple pages

## Verified

```bash
# correct id — works
runpodctl pod create --template-id runpod-torch-v21 --gpu-id "NVIDIA GeForce RTX 4090"
# → created successfully, then deleted

# wrong id — fails
runpodctl pod create --template-id runpod-torch-v21 --gpu-id "NVIDIA RTX 4090"
# → "There are no longer any instances available"
```

## Test plan

- [x] Built binary, confirmed help text shows correct GPU ID and image
- [x] E2E tested pod create with correct GPU ID — pod created and deleted
- [x] Confirmed wrong GPU ID fails as expected


justinwlin (approved):
LGTM — GPU ID fix ( → ) and updated Docker image are correct. Good call including the regenerated docs too.

PR #261, opened 2026-03-23
fix: use flashBootType, not -fb

k-nox:
indicate an endpoint is flashboot enabled with flashBootType instead of appending -fb to the name


TimPietruskyRunPod (approved):
tested e2e against the live api: created a serverless template and two endpoints via the graphql `saveEndpoint` mutation — one with `flashBootType: "FLASHBOOT"` and one with `flashBootType: "OFF"`. both were accepted by the api and created successfully. all test resources cleaned up.

lgtm — clean fix replacing the `-fb` name suffix hack with the proper api field.

PR #262, opened 2026-03-25
feat: add hub command to browse and deploy from the runpod hub

TimPietruskyRunPod:
## Summary

- Adds `runpodctl hub` command group (list, search, get) to browse the RunPod Hub marketplace via the existing GraphQL `listings` API
- Adds `--hub-id` flag to `serverless create` as an alternative to `--template-id`, enabling one-command deployment from hub repos
- When `--hub-id` is used, the CLI resolves the listing, extracts the build image and config from the latest release, creates an inline template via the GraphQL `saveEndpoint` mutation (matching the web UI flow), and deploys the endpoint

Linear: [DR-1325](https://linear.app/runpod/issue/DR-1325/runpodctl-create-a-serverless-endpoint-from-a-hub-listing-eg-vllm)

### New commands

```bash
runpodctl hub list                              # top 10 by stars
runpodctl hub list --type SERVERLESS            # only serverless repos
runpodctl hub search vllm                       # search by name
runpodctl hub get runpod-workers/worker-vllm    # get by owner/name
runpodctl hub get <listing-id>                  # get by id
```

### Serverless create from hub

```bash
runpodctl hub search vllm
runpodctl serverless create --hub-id cm8h09d9n000008jvh2rqdsmb --name "my-vllm"
```

GPU IDs and container disk size are automatically pulled from the hub release config. Users can override with \`--gpu-id\`.

## Test plan

- [x] Unit tests pass (\`go test ./...\`)
- [x] E2E tests for hub list, search, get (9 tests)
- [x] E2E tests for serverless create validation (2 tests)
- [x] E2E test for serverless create from hub with cleanup (1 test)
- [x] All 12 new e2e tests pass against live API
- [x] Docs regenerated


max4c (approved):
Looks good — clean command structure, solid E2E coverage, and good UX touches (owner/name format, helpful error messages). A couple minor suggestions for follow-up: (1) document that --type filtering is client-side so limit may return fewer results than expected, (2) check for double error output from output.Error + RunE return, (3) consider resolving the hardcoded listing ID in TestCLI_ServerlessCreateFromHub dynamically. None are blockers.

TimPietruskyRunPod:
thanks for the review @max4c! addressed all three points in 351603f:

1. **--type filtering docs** — added note to both `hub list` and `hub search` flag descriptions: "applied client-side; --limit may return fewer results"

2. **double error output** — this is actually fine by design. `SilenceErrors = true` in `cmd/root.go:77` prevents cobra from re-printing the error. the `output.Error(err)` + `return err` pattern is intentional across the whole codebase — `output.Error` emits the structured json error, and the returned error is used for exit code only.

3. **hardcoded listing id** — `TestCLI_ServerlessCreateFromHub` now resolves the vllm listing id dynamically via `hub get runpod-workers/worker-vllm` instead of the hardcoded `cm8h09d9n000008jvh2rqdsmb`.

justinwlin:
One issue I noticed in the new `serverless create --hub-id` path: Hub release env config is parsed but not included in the inline template used to create the endpoint.

`release.Config` is unmarshaled into `HubReleaseConfig`, and that struct includes `Env []HubReleaseConfigEnv`. But when building `EndpointTemplateInput`, the PR currently sends:

```go
Env: []*api.PodEnvVar{},
```

So any env vars declared by the Hub listing are silently dropped. For listings that rely on env defaults or required inputs, the endpoint creation can still succeed, but the resulting worker may boot with missing configuration or behave differently from the Hub/UI deployment.

For example, a Hub release config could include something like:

```json
{
  "gpuIds": "NVIDIA GeForce RTX 4090",
  "containerDiskInGb": 20,
  "env": [
    {
      "key": "MODEL_NAME",
      "input": {
        "default": "meta-llama/Llama-3.1-8B-Instruct"
      }
    }
  ]
}
```

In that case, `runpodctl serverless create --hub-id <id>` would currently create the endpoint with the image/GPU/disk settings, but without `MODEL_NAME`. The API call may succeed, but the worker could start without knowing which model to load, or it could fall back to different behavior than the Hub listing intended.

Suggestion: translate `hubConfig.Env` entries with defaults into `[]*api.PodEnvVar` and pass them into the inline template. If an env entry is required but has no default, the CLI should fail with a clear message, or support user-provided env overrides before creating the endpoint.

TimPietruskyRunPod:
great catch @justinwlin — fixed in ef21ba5:

1. **hub env vars now included** — all `hubConfig.Env` entries are translated into `PodEnvVar` entries in the inline template, using defaults where available. previously these were silently dropped (`Env: []*api.PodEnvVar{}`).

2. **new `--env` flag** — users can override hub defaults or add new env vars: `--env MODEL_NAME=my-model --env CUSTOM_KEY=value`. user values take precedence over hub defaults.

3. **e2e verified** — created an endpoint from the vllm hub listing with `--env MODEL_NAME=meta-llama/Llama-3.1-8B-Instruct`, confirmed the template included all 37 env vars from the hub config with correct defaults + the override. cleaned up after.

justinwlin (approved):
Approved. I pushed a formatting-only commit (`gofmt`) to clear the style finding.

One non-blocking follow-up I would still consider: `serverless create --hub-id` currently accepts any Hub listing type. Since this command creates Serverless endpoints, it may be worth rejecting listings where `listing.Type != "SERVERLESS"` with a clear error, unless deploying POD listings through this path is intentionally supported.

justinwlin:
Non-blocking follow-up: `serverless create --hub-id` currently does not check `listing.Type`, so it can accept a Hub repo whose type is `POD`. Since this command creates a Serverless endpoint, I would consider rejecting non-`SERVERLESS` listings with a clear error, unless deploying POD listings through this path is intentionally supported.

TimPietruskyRunPod:
re: listing type check — tested this and creating a serverless endpoint from a POD-type hub listing works fine (the API accepts it). this matches the web UI behavior where you can deploy either type as serverless.

added in 3ea02d5:
- documented in help text and `--hub-id` flag description that both SERVERLESS and POD types are accepted
- also updated the runpodctl skill (runpod/skills#12) with the `--env` flag and cross-type docs

TimPietruskyRunPod:
all feedback addressed — summary of changes since approval:

- **env vars fix** (ef21ba5) — hub release env config now included in inline template; added `--env` flag for overrides
- **listing type docs** (3ea02d5) — tested POD-type listings, they work fine for serverless create; documented in help text
- **spelling** (e7c1563) — RunPod → Runpod in user-facing strings
- **skill updated** (runpod/skills#12) — `--env` flag, cross-type docs, spelling fix

merging both PRs. will tag a release after merge. thanks for the reviews @justinwlin @max4c!

PR #264, opened 2026-04-08
add ssh remove-key command

vavo:
# E-258: add ssh remove-key command

- resolves #258 by adding `runpodctl ssh remove-key` with `--name` and `--fingerprint`.
- expands ssh key management coverage via the existing graphql `updateUserSettings` path, including ambiguity handling for duplicate key names.
- updates ssh command docs and adds unit coverage for command wiring and graphql removal behavior.
- tests: full unit suite passed locally; live e2e is still blocked in this workspace because no runpod api key is configured.

Test plan:
- [x] go test ./...
- [x] go test -tags e2e ./e2e


vavo:
live verification completed with a real Runpod API key.

verified flow:
- build `runpodctl`
- `ssh add-key` with disposable key A
- `ssh add-key` with disposable key B
- `ssh list-keys` to resolve key B fingerprint
- `ssh remove-key --name` for key A
- `ssh remove-key --fingerprint` for key B
- `ssh list-keys` to confirm both keys were removed

result: pass

cleanup was performed immediately after verification.

TimPietruskyRunPod (commented):
thanks for this — the remove-key feature is useful and the approach (fetch all keys, filter, rewrite via `updateUserSettings`) is sound.

a few things to address before merge:

**double error output**
`runSSHRemoveKey` calls `output.Error(err)` and then returns the error, which cobra also prints. same for the client creation error. pick one — either `output.Error` + `return nil`, or just `return err` and let cobra handle it. this pattern exists in a few places in the PR.

**test flakiness risk**
`TestSSHRemoveKey_RequiresIdentifier` mutates package-level `sshKeyName`/`sshKeyFingerprint` without cleanup. if tests run in parallel this will flake. use `t.Cleanup` to restore values.

**e2e testing**
per project constraints, e2e verification against the live api is non-negotiable before merge. the checklist shows this wasn't run. please do a round-trip: add a test key, remove it by name, verify it's gone, then remove by fingerprint.

no overlap with any other open PRs — this can merge independently once the above is addressed.

vavo:
hey tim, thanks for the review. i went through your comments and addressed the follow-up items on this branch.

changes made:
- removed the duplicate error output in `ssh remove-key`
- added cleanup in `testsshremovekey_requiresidentifier` so the package-level selector vars are restored after the test

verification:
- `go test ./cmd ./internal/api`
- live e2e verification passed against the api:
  - added 2 disposable ssh keys
  - resolved their fingerprints via `ssh list-keys`
  - removed one key by `--name`
  - removed the other by `--fingerprint`
  - verified both keys were gone afterward
  - cleaned up immediately after verification

TimPietruskyRunPod (approved):
looks good — all prior feedback addressed, clean implementation with proper ambiguity handling and e2e verification

PR #265, opened 2026-04-08
add min-cuda-version support for pod create

### LINKED ISSUE #253: Support --min-cuda-version flag for pod create
## Problem

When creating a pod with `runpodctl pod create`, there's no way to specify the minimum CUDA version required by the container image. This leads to pods being assigned to machines with incompatible (older) NVIDIA drivers, causing container startup failures:

```
nvidia-container-cli: requirement error: unsatisfied condition: cuda>=12.6,
please update your driver to a newer version, or use an earlier cuda container
```

The only workaround is to manually pick GPU types known to have recent drivers (e.g., RTX 5090), which defeats the purpose of flexible GPU selection.

## Proposed Solution

Add a `--min-cuda-version` flag (or similar) to `runpodctl pod create`:

```bash
runpodctl pod create \
  --image runpod/worker-comfyui:5.7.1-base \
  --gpu-id "NVIDIA GeForce RTX 4090" \
  --min-cuda-version 12.6 \
  ...
```

This would filter out machines whose driver doesn't support the requested CUDA version, preventing the pod from being scheduled on incompatible hardware.

## Context

The RunPod GraphQL API already has some filtering capabilities. Exposing CUDA version filtering through `runpodctl` would make it much easier to work with newer CUDA-based images without having to guess which GPU types or datacenters have updated drivers.

This came up while building automated Dockerfile verification scripts that spin up pods programmatically — having to hardcode specific GPU types as a driver workaround is fragile and costly.

vavo:
# E-253: add min-cuda-version support for pod create

- resolves #253 by adding `--min-cuda-version` to `runpodctl pod create` for gpu pod placement.
- aligns the graphql create path with the current RunPod schema, which includes `minCudaVersion` on `GpuAvailabilityInput` and `PodFindAndDeployOnDemandInput`.
- adds unit coverage for command wiring, gpu-only validation, request building, and graphql serialization.
- updates the generated `pod create` docs.

reference:
- https://graphql-spec.runpod.io/#definition-GpuAvailabilityInput

Test plan:
- [x] go test ./...
- [ ] go test -tags e2e ./e2e


vavo:
live verification completed with a real Runpod API key.

verified flow:
- build `runpodctl`
- `gpu list` confirmed available gpu ids, including `NVIDIA GeForce RTX 4090`
- create disposable pod with:
  - `--image ubuntu:22.04`
  - `--gpu-id 'NVIDIA GeForce RTX 4090'`
  - `--min-cuda-version 12.6`
  - `--name codex-issue-253-probe`
- create succeeded and returned pod id `nuyi29y19kylna`
- immediately deleted the disposable pod with `runpodctl pod delete nuyi29y19kylna`

result: pass

cleanup completed successfully.

TimPietruskyRunPod:
thanks for working on this. we've since landed a superset of this change in #271 which adds `--min-cuda-version` to both pod create AND serverless create, wires it into the REST path (so CPU pods don't silently ignore it), and was e2e verified against the live api.

a few issues we noticed in this PR:
- the REST path (`createPodREST`) never sets `MinCudaVersion`, so CPU pods silently drop the flag
- e2e testing wasn't run
- the `buildCreatePodGQLRequest` refactor manually saves/restores 14 package-level vars in tests — brittle

closing in favor of #271. appreciate the work — the core approach (sending `minCudaVersion` to GraphQL) is identical to what we ended up with.

TimPietruskyRunPod:
closing in favor of #271 (superset)

Issue #252, opened 2026-03-06
feat: add missing flags for serverless/template management


TimPietruskyRunPod:
## Problem

Several critical serverless and template management operations are missing from the CLI, forcing users to fall back to the web UI or raw GraphQL API calls.

### Missing flags

**`runpodctl serverless create`**
- `--network-volume-id` — no way to attach a network volume when creating an endpoint

**`runpodctl serverless update`**
- `--template-id` — can't change the template on an existing endpoint

**`runpodctl template create`**
- `--volume-mount-path` is accepted but ignored (always returns `/workspace`)

**`runpodctl template update`**
- `--container-disk-in-gb` — can't update container disk size on an existing template

### Impact

When deploying a custom Docker image to serverless (e.g. a ComfyUI worker that needs a network volume for models and >5GB container disk), none of these operations can be done end-to-end via the CLI. Users must switch to the web UI mid-workflow.

### Expected behavior

All CRUD operations available in the web UI should be available via `runpodctl`.

justinwlin:
All four gaps are closed on `main` (v2.8.0):

- `serverless create --network-volume-id` — `cmd/serverless/create.go:95` (plus `--network-volume-ids` for multi-region; the two are mutually exclusive)
- `serverless update --template-id` — `cmd/serverless/update.go:49`
- `template create --volume-mount-path` — now actually sent instead of being ignored, `cmd/template/create.go:92`
- `template update --container-disk-in-gb` — `cmd/template/update.go:42`

The ComfyUI-style workflow (custom image + network volume + >5GB container disk) is now doable end-to-end from the CLI. Closing.

PR #266, opened 2026-04-08
add missing serverless and template update flags

vavo:
# E-252: add missing serverless and template update flags

- resolves the remaining cli gaps from #252 by adding `runpodctl serverless update --template-id` and `runpodctl template update --container-disk-in-gb`.
- updates endpoint template swapping through the existing graphql mutation path and extends template update requests to include container disk size.
- adds unit coverage for the new flags and request paths, and updates generated docs for the affected commands.
- note: the other two report items in #252 are already present/working on current `main`:
  - `runpodctl serverless create --network-volume-id` already exists
  - `runpodctl template create --volume-mount-path` already works; i live-verified `/models` persisted on create/get

Test plan:
- [x] go test ./...
- [ ] go test -tags e2e ./e2e
- [x] live verify `template update --container-disk-in-gb` on a disposable template, then delete it
- [x] live verify `serverless update --template-id` on a disposable endpoint with two disposable serverless templates, then delete all resources


TimPietruskyRunPod (commented):
thanks for this — the `serverless update --template-id` and `template update --container-disk-in-gb` flags are genuinely useful additions. this is complementary to #271 (which adds flags to `create` commands, not `update`), so no overlap.

a few things to address:

**partial update on failure**
if the REST update succeeds but the GraphQL `--template-id` call fails, the endpoint is left partially updated with no rollback or warning. at minimum, print a warning to stderr saying the REST fields were updated but the template swap failed.

**container-disk-in-gb zero-value problem**
the `ContainerDiskInGb` field uses `int` with `omitempty` in the JSON tag. setting it to `0` would silently omit it from the request. if `0` is a valid value the user might want to send, this is a bug. consider using `*int` or removing `omitempty`.

**double error reporting**
`runUpdate` calls `output.Error(err)` and then returns `fmt.Errorf(...)`. cobra will also handle the returned error, so the user sees the error twice. pick one path.

**e2e testing**
please verify against the live api — create an endpoint, update its template-id, confirm the change in the response, clean up.

this can merge alongside #271 with minor conflict resolution on `endpoints.go` whitespace. looking forward to it.

vavo:
Hi Tim, thanks for the review, I'll fix the PR when I get a chance.

regarding e2e tests - for some reason I've though this is something you run as part of the review. From now on I'll include them as part of the PRs.

vavo:
Hi Tim, went through the follow-up items and updated the branch. Let me know if I forgot something.

changes made:
- added a stderr warning for the partial-update case where the rest endpoint update succeeds but the graphql template swap fails
- changed template update request handling so `container-disk-in-gb` is sent explicitly when set, including `0`
- removed the duplicate local error output in the updated commands so errors are surfaced once

verification:
- `go test ./cmd/serverless ./cmd/template ./internal/api`
- added unit coverage for graphql error handling, the partial-update warning path, and `container-disk-in-gb=0`
- live e2e verification passed against the api:
  - created a disposable serverless template and updated `container-disk-in-gb` from `20` to `21`
  - created a second disposable serverless template
  - created a disposable cpu endpoint from the first template
  - updated the endpoint with `--template-id` to the second template
  - confirmed the endpoint reported the new template id afterward
  - deleted the endpoint and both templates, then verified they were gone

TimPietruskyRunPod (commented):
thanks for addressing all the feedback — the partial-update warning, the `*int` fix for `container-disk-in-gb`, and removing the duplicate error output all look good.

one last thing: if the user runs `serverless update <id>` without passing any flags, the command silently succeeds and just prints the current endpoint state. it would be better to return an error like `"no update flags specified"` so the user knows they forgot to specify what to update. same applies to `template update`.

once that's in, this is good to go.

TimPietruskyRunPod (approved):
looks good — serverless template swap and container disk flag are solid, tests pass, conflicts resolved

lavanya-gunreddi:
@vavo @TimPietruskyRunPod https://app.gopromptless.ai/suggestions/9c59c4de-7ddf-4709-ba54-4701dbf83f59 - Pushing these changes suggested by Promptless for this Github change.

PR #276, opened 2026-05-07
feat: CE-1579: Add support for deploying serverless with models

brosenpod:
# CE-1579: Add support for deploying serverless with models

`runpodctl` was not previously able to deploy a serverless endpoint with a model attached. Normal serverless deploy uses REST API, but REST API also does not support this. When `runpodctl` is asked to deploy a serverless endpoint with a model it will instead talk to GraphQL for the deploy. While this bifurcation is not optimal, we already do this when deploying from the Hub.


cmd/pod/list.go
```diff
@@ -21,15 +21,15 @@ var listCmd = &cobra.Command{
...
-	GpuCount      int         `json:"gpuCount"`
-	VolumeInGb    int         `json:"volumeInGb"`
-	CostPerHr     float64     `json:"costPerHr,omitempty"`
-	CreatedAt     string      `json:"createdAt,omitempty"`
+	ID            string  `json:"id"`
+	Name          string  `json:"name"`
+	DesiredStatus string  `json:"desiredStatus"`
+	ImageName     string  `json:"imageName"`
+	GpuID         string  `json:"gpuId,omitempty"`
+	GpuCount      int     `json:"gpuCount"`
```
TimPietruskyRunPod: unrelated cosmetic change — please drop this file from the PR

internal/api/endpoints.go
```diff
@@ -146,7 +147,7 @@ func (c *Client) DeleteEndpoint(endpointID string) error {
 // EndpointCreateGQLInput is the input for creating an endpoint via GraphQL
 // Used when hubReleaseId is needed (REST API doesn't support it)
 type EndpointCreateGQLInput struct {
-	Name            string                 `json:"name"`
+	Name            string                 `json:"name,omitempty"`
```
TimPietruskyRunPod, line 201: making `Name` omitempty also affects the existing hub-id path — if someone passes an empty name, it won't be sent to the API. in practice the hub path always sets a name, but this is a silent behavior change to an existing code path. please revert this to non-omitempty and handle the empty-name case explicitly in your new code path if needed.

cmd/serverless/create.go
```diff
@@ -24,6 +24,9 @@ examples:
   # create from a template
   runpodctl serverless create --template-id <id> --gpu-id "NVIDIA GeForce RTX 4090"
 
+  # create from a template and attach a model
```
TimPietruskyRunPod, line 27: `https://local/...` doesn't tell users what an actual model reference looks like. please use a realistic example so it's clear what format to pass.

cmd/serverless/create.go
```diff
@@ -211,6 +223,18 @@ func runCreate(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
+			output.Error(err)
+			return fmt.Errorf("failed to create endpoint: %w", err)
+		}
+
+		format := output.ParseFormat(cmd.Flag("output").Value.String())
+		return output.Print(endpoint, &output.Config{Format: format})
+	}
+
 	endpoint, err := client.CreateEndpoint(req)
```
TimPietruskyRunPod, line 285: `ComputeType` from the request is not passed into the GQL input. the validation above rejects non-GPU, but it's still not explicitly sent — please confirm the API defaults to GPU, or add it to the GQL input.

TimPietruskyRunPod (dismissed):
nice approach — reusing the existing GQL path for model references makes sense. a few things to address before this is ready.

internal/api/endpoints.go
TimPietruskyRunPod: actually disregard this — the omitempty is needed for the model-reference path where `--name` is optional and the API should auto-generate one. the hub-id path always sets a name so it's not affected. this is fine as-is.

murat-runpod (approved):

lavanya-gunreddi:
@promptless Do you have the latest changes updated to the documentation?

PR #277, opened 2026-05-07
fix: Bump versions for vuln remediation

brosenpod:
# fix: Bump versions for vuln remediation

Fix vulnerabilities and issues found from `govulncheck ./...`


k-nox:
probably also need to bump it in github workflows: https://github.com/runpod/runpodctl/blob/d34f0838afc2f5cc6d0c4ac79501a119b70e78c2/.github/workflows/ci.yml#L11 and https://github.com/runpod/runpodctl/blob/d34f0838afc2f5cc6d0c4ac79501a119b70e78c2/.github/workflows/release.yml#L37
you can set it to use the go.mod for version detection with `go-version-file: go.mod` i think?

brosenpod:
@k-nox Good catch! I have updated those as well.

k-nox (approved):

PR #278, opened 2026-05-11
feat: CE-1584: Show model version hash in model list

brosenpod:
# CE-1584: Show model version hash in model list

Adds a Version Hash column to runpodctl model list so users can copy the model version hash needed for serverless endpoint model references.

The list output now uses the first non-empty version hash returned for each model and falls back to "-" when no hash is available. Also adds focused unit coverage for hash selection and updates the runpodctl skill note to mention the new output.

### DIFF
--- a/cmd/model/getModels.go
+++ b/cmd/model/getModels.go
@@ -76,12 +76,13 @@ func runModelList(cmd *cobra.Command, args []string) {
 func displayModels(models []*api.Model) {
 	w := tabwriter.NewWriter(os.Stdout, 0, 8, 2, ' ', 0)
 
-	fmt.Fprintln(w, "ID\tProvider\tName\tOwner\tStatus\tCreated At\tUpdated At")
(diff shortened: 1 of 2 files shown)


cmd/model/getModels.go
```diff
@@ -95,6 +96,20 @@ func displayModels(models []*api.Model) {
 	}
 }
 
+func modelVersionHash(model *api.Model) string {
+	if model == nil {
+		return ""
+	}
+
+	for _, version := range model.Versions {
```
javier-runpod, line 104: Can we just do the following since empty string is a valid string?

```go
for _, version := range model.Versions {
		if version != nil {
			return strings.TrimSpace(version.Hash)
		}
	}
```

javier-runpod (approved):

cmd/model/getModels.go
brosenpod: We want to make sure that we return a usable hash, so an empty string is also not helpful.

PR #280, opened 2026-05-26
feat: add agent source tracking to user-agent header

KAJdev:
Appends coding agent source tags to the User-Agent header on all outgoing API requests when the corresponding environment variable is set (e.g. `CLAUDECODE=1` adds `(via claude-code)`). Covers both the REST client and GraphQL client paths.

CON-85

### DIFF
--- a/api/query.go
+++ b/api/query.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/runpod/runpodctl/internal/agent"
 	"github.com/spf13/viper"
 )
 
@@ -58,8 +59,8 @@ func Query(input Input) (res *http.Response, err error) {
 
 	sanitizedVersion := strings.TrimRight(Version, "\r\n")
 	userAgent := "RunPod-CLI/" + sanitizedVersion + " (" + runtime.GOOS + "; " + runtime.GOARCH + ")"
-	if os.Getenv("CLAUDECODE") == "1" {
-		userAgent += " (via claude-code)"
+	if a := agent.Detect(); a != "" {
+		userAgent += " (via " + a + ")"
 	}
 
 	req.Header.Add("Content-Type", "application/json")
--- a/api/query.go
+++ b/api/query.go
@@ -58,6 +58,9 @@ func Query(input Input) (res *http.Response, err error) {
 
 	sanitizedVersion := strings.TrimRight(Version, "\r\n")
 	userAgent := "RunPod-CLI/" + sanitizedVersion + " (" + runtime.GOOS + "; " + runtime.GOARCH + ")"
+	if os.Getenv("CLAUDECODE") == "1" {
+		userAgent += " (via claude-code)"
+	}
 
 	req.Header.Add("Content-Type", "application/json")
 	req.Header.Set("User-Agent", userAgent)
--- a/api/query.go
+++ b/api/query.go
@@ -58,10 +58,7 @@ func Query(input Input) (res *http.Response, err error) {
 	}
 
 	sanitizedVersion := strings.TrimRight(Version, "\r\n")
-	userAgent := "RunPod-CLI/" + sanitizedVersion + " (" + runtime.GOOS + "; " + runtime.GOARCH + ")"
-	if a := agent.Detect(); a != "" {
-		userAgent += " (via " + a + ")"
-	}
+	userAgent := "RunPod-CLI/" + sanitizedVersion + " (" + runtime.GOOS + "; " + runtime.GOARCH + ")" + agent.Suffix()
 
 	req.Header.Add("Content-Type", "application/json")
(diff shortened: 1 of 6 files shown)


internal/agent/agent.go
```diff
@@ -0,0 +1,79 @@
...
+	{id: "replit", envVars: []string{"REPL_ID"}},
+	{id: "trae", envVars: []string{"TRAE_AI_SHELL_ID"}},
+	{id: "zed", envVars: []string{"ZED_TERM"}},
+	{id: "cursor-cli", envVars: []string{"CURSOR_AGENT"}},
+	{id: "cursor", envVars: []string{"CURSOR_TRACE_ID"}},
+}
+
+// standardEnvVars are generic variables any tool can set to identify itself.
+// When set, the value is used directly as the agent id.
+var standardEnvVars = []string{"AI_AGENT", "AGENT"}
```
TimPietruskyRunPod: **medium — drop bare `AGENT` or namespace it.** `AGENT` is a very common variable (CI runners, tmux/ssh setups, ad-hoc scripts), and its raw value is used directly as the agent id. Unrelated environments will silently tag traffic with arbitrary strings, polluting the analytics this feature exists to produce. `AI_AGENT` is specific enough to keep; I'd remove bare `AGENT` or gate it behind a `RUNPOD_`-namespaced opt-in.

internal/agent/agent.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, h := range harnesses {
+		for _, env := range h.envVars {
+			if os.Getenv(env) != "" {
+				return h.id
+			}
+		}
+	}
+	for _, env := range standardEnvVars {
+		if v := os.Getenv(env); v != "" {
+			return v
```
TimPietruskyRunPod, line 90: **low — sanitize the fallback value before it reaches the header.** This value is concatenated straight into the User-Agent. It isn't header injection (net/http rejects CRLF at write time), but a value with spaces/junk produces a malformed UA. Suggest `strings.TrimSpace` plus a charset/length clamp (e.g. keep `[A-Za-z0-9._-]`, cap ~64) here before returning.

api/query.go
```diff
@@ -58,6 +59,9 @@ func Query(input Input) (res *http.Response, err error) {
 
 	sanitizedVersion := strings.TrimRight(Version, "\r\n")
 	userAgent := "RunPod-CLI/" + sanitizedVersion + " (" + runtime.GOOS + "; " + runtime.GOARCH + ")"
+	if a := agent.Detect(); a != "" {
```
TimPietruskyRunPod: **nit (non-blocking)** — the UA string is now assembled in two places: `internal/api.buildUserAgent()` (`runpod-cli/...`) and here inline (`RunPod-CLI/...`, different casing, empty version in non-release builds). The package split makes sharing awkward and the casing is pre-existing, so fine to leave — but worth a follow-up to centralize so the `(via ...)` tag can't drift between paths.

TimPietruskyRunPod (changes requested):
Clean, well-tested feature and verified live against production across the REST, GraphQL, and legacy paths — every path emits the `(via <agent>)` tag and the API accepts it. Rebased onto main is green (`go build`/`vet`/`test ./...`).

Requesting changes on two points before merge, both small:

1. **Bare `AGENT` fallback risks false attribution** (medium) — it's a widely-used variable, so unrelated environments will silently tag traffic with arbitrary strings and pollute the very analytics this feature produces.
2. **Env value flows unsanitized into the User-Agent** (low) — not injectable (Go rejects CRLF at write time), but spaces/junk yield a malformed UA. A trim + charset/length clamp would harden it.

Non-blocking nit left inline on the duplicated UA assembly. Happy to fold the fixes in directly if you'd prefer.

TimPietruskyRunPod:
Pushed `30992cc` addressing the two review points:

- **dropped bare `AGENT`** — only `AI_AGENT` is honored now, so common environments (ci runners, shell setups) can't false-attribute traffic.
- **sanitized the standard value** — trimmed + clamped to `[A-Za-z0-9._-]`, 64-rune cap, so an arbitrary env value can't produce a malformed user-agent.

added unit tests for both (bare-AGENT-ignored, injection-stripped, length cap). verified live against production: `gpu list` etc. still emit `(via claude-code)`, an injected `AI_AGENT=$'evil agent\\r\\nX-Injected: 1'` lands as `(via evilagentX-Injected1)`, and a bare `AGENT` produces no tag. the (non-blocking) duplicated UA-assembly nit in `api/query.go` is left for a follow-up.

TimPietruskyRunPod (approved):
all review points addressed in 30992cc (drop bare AGENT + sanitize standard value) and 84c3c92 (centralize the (via <id>) suffix). build/vet/test green; verified live against production across rest, graphql, and legacy paths. approving.

PR #281, opened 2026-05-26
fix: add modelReferences to endpoint create/update and document in runpod.toml template

jebenexer:
## Summary

Adds support for `modelReferences` in the endpoint create/update flow.

### Changes

1. **`api/endpoint.go`** — Added `ModelReferences []string` field to `CreateEndpointInput`; fixed missing `json:"id"` tag on `Endpoint.Id`; added `modelReferences` to `GetEndpoints` GraphQL query; added `UpdateEndpointModel()` function that uses `saveEndpoint` mutation to update models on existing endpoints

2. **`cmd/project/functions.go`** — `deployProject` reads `model_refs` from endpoint config and passes it as `ModelReferences` to `CreateEndpointInput`

3. **`cmd/project/tomlBuilder.go`** — Documents the new `model_refs` config key in the `[endpoint]` section of the `runpod.toml` template

4. **`cmd/endpoint/update_model.go`** (new) — Adds `runpodctl endpoint model <endpoint-id> <model-ref> [model-ref...]` CLI command to set model references on an existing endpoint

5. **`cmd/root.go`** — Registers the new endpoint command group

### Related

- Backend PR: https://github.com/runpod/RunPod/pull/4678
- Closes SLS-94

### DIFF
--- a/api/endpoint.go
+++ b/api/endpoint.go
@@ -33,12 +33,13 @@ type CreateEndpointInput struct {
 	WorkersMin      int    `json:"workersMin"`
 	WorkersMax      int    `json:"workersMax"`
 	FlashBootType   string `json:"flashBootType"`
+	ModelReferences []string `json:"modelReferences"`
 }
 
 // there are many more fields in the result of the query but I just care about these for CLI port
 type Endpoint struct {
 	Name string `json:"name"`
-	Id   string
+	Id   string `json:"id"`
 }
 type EndpointOut struct {
 	Data   *EndpointData   `json:"data"`
@@ -225,6 +226,53 @@ func UpdateEndpointTemplate(endpointId string, templateId string) (err error) {
 	return
 }
 
+func UpdateEndpointModel(endpointId string, endpointName string, modelReferences []string) (err error) {
+	input := Input{
+		Query: `
+		mutation saveEndpoint($input: EndpointInput!) {
+			saveEndpoint(input: $input) {
+			  id
+			  modelReferences
+			}
+		  }
+		`,
+		Variables: map[string]interface{}{"input": map[string]interface{}{
+			"id":              endpointId,
+			"name":            endpointName,
+			"modelReferences": modelReferences,
+		}},
+	}
+	res, err := Query(input)
+	if err != nil {
+		return
+	}
+	defer res.Body.Close()
+	rawData, err := io.ReadAll(res.Body)
+	if err != nil {
+		return
+	}
+	if res.StatusCode != 200 {
+		err = fmt.Errorf("statuscode %d: %s", res.StatusCode, string(rawData))
+		return
+	}
+	data := make(map[string]interface{})
+	if err = json.Unmarshal(rawData, &data); err != nil {
+		return
+	}
+	gqlErrors, ok := data["errors"].([]interface{})
+	if ok && len(gqlErrors) > 0 {
+		firstErr, _ := gqlErrors[0].(map[string]interface{})
+		err = errors.New(firstErr["message"].(string))
+		return
+	}
+	gqldata, ok := data["data"].(map[string]interface{})
+	if !ok || gqldata == nil {
+		err = fmt.Errorf("data is nil: %s", string(rawData))
+		return
+	}
+	return
+}
+
 func GetEndpoints() (endpoints []*Endpoint, err error) {
 	input := Input{
 		Query: `
@@ -248,6 +296,7 @@ func GetEndpoints() (endpoints []*Endpoint, err error) {
 				workersMin
 				workersStandby
 				gpuCount
+				modelReferences
 				env {
 				  key
 				  value
--- a/api/endpoint.go
+++ b/api/endpoint.go
@@ -33,12 +33,13 @@ type CreateEndpointInput struct {
 	WorkersMin      int    `json:"workersMin"`
 	WorkersMax      int    `json:"workersMax"`
 	FlashBootType   string `json:"flashBootType"`
+	ModelReferences []string `json:"modelReferences"`
 }
 
 // there are many more fields in the result of the query but I just care about these for CLI port
 type Endpoint struct {
 	Name string `json:"name"`
-	Id   string
+	Id   string `json:"id"`
 }
 type EndpointOut struct {
 	Data   *EndpointData   `json:"data"`
--- a/internal/api/endpoints.go
+++ b/internal/api/endpoints.go
@@ -188,6 +188,106 @@ func (c *Client) DeleteEndpoint(endpointID string) error {
 	return err
 }
 
+// UpdateEndpointModels sets the model references on an existing endpoint via
+// saveEndpoint. The full current config is round-tripped so that only
+// modelReferences changes — saveEndpoint is a full replace and omitting fields
+// would reset them to server defaults. Pass nil or an empty slice to clear all
+// model references.
+func (c *Client) UpdateEndpointModels(endpointID string, modelRefs []string) (*Endpoint, error) {
+	endpoint, err := c.GetEndpoint(endpointID, false, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch endpoint: %w", err)
+	}
+
+	if modelRefs == nil {
+		modelRefs = []string{}
+	}
+
+	// saveEndpoint expects networkVolumeIds as [{networkVolumeId}] objects; the
+	// REST read shape uses bare id strings which UnmarshalJSON normalises into
+	// EndpointNetworkVolume — convert to the GraphQL write shape here.
+	nvIDs := make([]NetworkVolumeIDInput, len(endpoint.NetworkVolumeIDs))
+	for i, nv := range endpoint.NetworkVolumeIDs {
+		nvIDs[i] = NetworkVolumeIDInput{NetworkVolumeID: nv.NetworkVolumeID}
+	}
+
+	query := `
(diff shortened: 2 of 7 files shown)


TimPietruskyRunPod (changes requested):
the `modelReferences` plumbing here is solid — `CreateEndpointInput`, the `GetEndpoints` query addition, `UpdateEndpointModel()` via `saveEndpoint`, and the `runpod.toml` `model_refs` key are all the right pieces. requesting changes on the **CLI surface** only.

**don't introduce a new top-level `endpoint` group — use the existing `serverless` group.**

we already have `serverless` (`create`/`delete`/`get`/`list`/`update`), and `serverless create --model-reference` already landed in #276. adding `endpoint model` alongside it gives two command namespaces for the same resource, and `model` would be the only thing under `endpoint` — so the surface fragments and `--model-reference` (create) vs `endpoint model` (update) becomes inconsistent for the same concept.

`endpoint model <id> [refs...] [--clear]` is functionally just a `serverless update` that only touches model refs. fold it into the command that already exists:

- set/replace: `runpodctl serverless update <id> --model-reference a --model-reference b`
- remove all:  `runpodctl serverless update <id> --clear-models`

concretely:
- add a repeatable `--model-reference` (`StringArrayVar`) and a `--clear-models` bool to `cmd/serverless/update.go`, matching the `--model-reference` verb already on `serverless create`
- on update, set `modelReferences` on the `saveEndpoint` input: the provided refs, or `[]` when `--clear-models` is passed (same empty-list-to-clear behavior the current `--clear` already relies on)
- keep all the api / `runpod.toml` plumbing from this PR as-is — only the command placement changes
- drop `cmd/endpoint/` (and revert the docs to extend the serverless reference page instead of a new endpoint page)

the only thing that would justify a new group is a deliberate rename `serverless` → `endpoint` (the API/console domain term is "Endpoint") with `serverless` kept as a hidden alias — but that's a bigger decision and should be its own PR with a deprecation plan, not introduced as a side effect of adding model refs.

tl;dr: plumbing 👍 — move `endpoint model` → `serverless update --model-reference / --clear-models` and drop the new top-level group.

TimPietruskyRunPod:
to be clear on scope — the **read-side additions are great, keep them**:

- adding `modelReferences` to the `GetEndpoints()` query selection so cached models surface in `serverless get` / `serverless list` 👍
- the `Endpoint.ModelReferences` field + the endpoint-id json unmarshalling fix 👍
- `model_refs` in `runpod.toml` + the project-deploy passthrough 👍

the **only** change requested is the write-side command placement: drop the new top-level `endpoint` group (`cmd/endpoint/`) and expose the update via `serverless update --model-reference / --clear-models` instead. everything else in the PR stays.

so: read extension to `serverless get` = 🔥, just not the `endpoint model` command.

jebenexer:
done — dropped `cmd/endpoint/` entirely and folded the write-side into `serverless update`:

- `runpodctl serverless update <id> --model-reference <ref> [--model-reference <ref2>...]` — set/replace
- `runpodctl serverless update <id> --clear-models` — remove all

the two are mutually exclusive (validated at runtime). the `api/endpoint.go` plumbing, `runpod.toml` `model_refs`, `GetEndpoints` query addition, and `ModelReferences` field are all unchanged. `UpdateEndpointModels` lands in `internal/api/endpoints.go` via `saveEndpoint` graphql. all unit tests pass, e2e verified against the live api.

jebenexer:
Done — `cmd/endpoint/` is gone and the new top-level `endpoint` group with it. `--model-reference` (repeatable, replaces existing) and `--clear-models` are now flags on `serverless update`, matching the verb already on `serverless create`. The `model_refs` TOML plumbing and the `UpdateEndpointModels` API helper are unchanged. History is squashed to a single commit on top of current main.

internal/api/endpoints.go
```diff
@@ -169,6 +169,67 @@ func (c *Client) DeleteEndpoint(endpointID string) error {
 	return err
 }
 
+// UpdateEndpointModels sets the model references on an existing endpoint via
+// saveEndpoint. The server requires the endpoint name alongside the id, so the
+// current endpoint is fetched first. Pass nil or an empty slice to clear all
+// model references.
+func (c *Client) UpdateEndpointModels(endpointID string, modelRefs []string) (*Endpoint, error) {
```
TimPietruskyRunPod, line 196: **blocking — this resets the endpoint's other fields to defaults.**

`saveEndpoint` is a full replace, not a merge. sending only `id` + `name` + `modelReferences` here causes the backend to reset every omitted field. confirmed live on this branch: an endpoint created with `idleTimeout=42, scalerValue=9, workersMax=5` came back as `10 / 4 / 3` after a model-only `serverless update`.

the function already fetches the endpoint (for `endpoint.Name`) — extend that to round-trip the rest of the config (`workersMin/workersMax/idleTimeout/scalerType/scalerValue/gpuIds/templateId/...`) into the `input`, so only `modelReferences` actually changes. note `GetEndpoint` here is the REST shape — confirm it returns all the fields `saveEndpoint` needs, or fetch them via graphql.

api/endpoint.go
```diff
@@ -225,6 +226,53 @@ func UpdateEndpointTemplate(endpointId string, templateId string) (err error) {
 	return
 }
 
+func UpdateEndpointModel(endpointId string, endpointName string, modelReferences []string) (err error) {
```
TimPietruskyRunPod: dead code — `UpdateEndpointModel` (singular) is never called anywhere on the branch. the live command uses `internal/api.UpdateEndpointModels`. leftover from the dropped `cmd/endpoint/` approach; please remove.

api/endpoint.go
```diff
@@ -248,6 +296,7 @@ func GetEndpoints() (endpoints []*Endpoint, err error) {
 				workersMin
 				workersStandby
 				gpuCount
+				modelReferences
```
TimPietruskyRunPod: this `modelReferences` selection is a no-op: the `Endpoint` struct in this package only has `Name`/`Id` (no `ModelReferences` field), so the value is fetched and discarded. the only caller is `cmd/project/functions.go` (lookup by name), and `serverless get`/`list` use `internal/api` instead. either drop this line or add the field where it's actually consumed.

TimPietruskyRunPod (dismissed):
## re-review — command placement is fixed, but the update path ships data loss

thanks for folding the write-side into `serverless update --model-reference / --clear-models` and dropping the top-level `endpoint` group — that part is exactly right, and the create/toml plumbing is good.

**blocking:** `serverless update --model-reference` (and `--clear-models`) silently resets the rest of the endpoint's config to defaults. `saveEndpoint` is a full **replace**, not a merge, and `UpdateEndpointModels` only sends `id` + `name` + `modelReferences`, so every other field reverts.

verified end-to-end against the live api on this branch:
- created an endpoint with `idleTimeout=42, scalerValue=9, workersMax=5`
- ran `serverless update <id> --model-reference https://huggingface.co/Qwen/Qwen2.5-0.5B-Instruct:main`
- re-fetched: `idleTimeout=10, scalerValue=4, workersMax=3` — all reset to defaults

this is net-new in this PR: on `main`, `saveEndpoint` is only ever called for *create* (no `id`), and the existing `serverless update` uses the REST PATCH (`UpdateEndpoint`) + the dedicated `updateEndpointTemplate` mutation for partial updates — neither touches `saveEndpoint`. the safe partial-update path was already established; the model-update code just didn't follow it. it also breaks the combined case (`--workers-min 5 --model-reference X`): the REST patch sets workers, then `saveEndpoint` immediately wipes them.

the schema confirms only `name` is required on `EndpointInput`, so the partial mutation succeeds silently with no error — which is what makes this a quiet footgun rather than a loud failure.

fix options:
1. round-trip the full current config — `UpdateEndpointModels` already fetches the endpoint for its name; fetch the rest (`workersMin/workersMax/idleTimeout/scalerType/scalerValue/gpuIds/templateId/...`) and echo it back alongside `modelReferences`, or
2. use/ask backend for a dedicated partial mutation (the way `updateEndpointTemplate` exists separately is the hint that `saveEndpoint` isn't safe for partial updates).

plus two cleanups inline. happy to push the round-trip fix if useful.

internal/api/endpoints.go
jebenexer: Fixed. `UpdateEndpointModels` now round-trips the full current config via the initial `GetEndpoint` REST fetch: `templateId`, `gpuIds`, `gpuCount`, `instanceIds`, `workersMin`, `workersMax`, `locations`, `networkVolumeId`, `networkVolumeIds`, `idleTimeout`, `scalerType`, `scalerValue`, `executionTimeoutMs`, `minCudaVersion`, `flashBootType` are all echoed back alongside `modelReferences`. Added `TestUpdateEndpointModels_RoundTripsConfig` (unit) and updated `TestRunUpdate_ModelReferences` (command) to serve custom config values and assert they're preserved. Also handles the bare-string `networkVolumeIds` REST read shape → `[{networkVolumeId}]` write shape conversion, using the `UnmarshalJSON` from #292.

api/endpoint.go
jebenexer: Removed.

api/endpoint.go
jebenexer: Dropped the selection. `api.Endpoint` only has `Name`/`Id` so the field was fetched and discarded. The one caller (`cmd/project/functions.go`) only needs the name for a lookup.

brosenpod (approved):

PR #284, opened 2026-06-10
feat: unify serverless create on graphql saveEndpoint

TimPietruskyRunPod:
> **note:** this branch is based on #276 and includes its commit, so the diff below shows #276's work + our changes combined. it effectively supersedes #276 — if this merges, #276 can be closed (or cherry-pick commit `38abee0` onto the #276 branch to keep that as the landing PR).

builds on top of #276. unifies all of `serverless create` onto the graphql `saveEndpoint` mutation instead of routing some paths through rest and others through graphql.

## why

#276 added `--model-reference`, which forces the graphql path. but the bifurcation meant the path you hit (template → rest, model/hub → graphql) silently decided which flags actually worked. on the graphql paths, `--idle-timeout`, `--scale-by/--scale-threshold`, `--execution-timeout`, `--min-cuda-version`, `--network-volume-ids`, and `--flash-boot` were silently dropped. the web console uses `saveEndpoint` for everything, and `EndpointInput` is a superset of the rest fields — so unifying on it is the way to get every capability on one path with nothing dropped.

## what changed

- **single path**: template, hub, model-reference, cpu, and gpu all go through `saveEndpoint`. removed the now-unused rest `CreateEndpoint`.
- **no silent drops**: every flag maps to `EndpointInput`. conflicting flags now error instead of being ignored (e.g. `--network-volume-id` + `--network-volume-ids`).
- **gpu id translation**: `saveEndpoint.gpuIds` wants gpu *pool* ids (`ADA_24`), not the gpu *type* ids (`NVIDIA A40`) that `gpu list` / `--gpu-id` use. `--gpu-id` is now translated via `serverlessGpuPools` (type id or pool id both work).
- **cpu**: selected via `instanceIds` (new `--instance-id`, default `cpu3g-4-16`); `saveEndpoint` has no `computeType`.
- **flash boot**: maps to the `flashBootType` enum, so `--flash-boot=false` is honored without the old post-create rest patch.
- **multi-region volumes**: `--network-volume-ids` sent as `[{networkVolumeId}]`.
- **name**: `saveEndpoint` requires a name (`String!`, min 3) and never auto-generates one, so a default is generated client-side when `--name` is omitted.
- **output**: create response now echoes `instanceIds` and `networkVolumeIds` so cpu/multi-region endpoints are visible.

## testing

- unit tests updated (table-driven validations, serialization); `go test ./...` passes.
- e2e against the live api (all resources cleaned up): gpu type-id→pool-id translation (`NVIDIA GeForce RTX 4090`→`ADA_24`, `NVIDIA A40`→`AMPERE_48`), cpu default + explicit instance, flags-take-effect verified server-side via graphql (idleTimeout/scaler/executionTimeoutMs/flashBootType=OFF), single + repeated model references, name auto-generation, data-center-ids, single + multi-region volumes, hub deploy, and all 6 validation/error paths.

docs and AGENTS.md updated.


justinwlin:
## Review — unify serverless create on `saveEndpoint`

Good direction (single path, kills the silent-routing flag drops). Builds clean, 61 tests pass. Findings below by severity — flagging one likely-breaking case to verify first.

### 🔴 High — verify first

**H1. Hub-config `gpuIds` fallback sent untranslated to `saveEndpoint`** — `cmd/serverless/create.go` (~L268)

**Background — two different GPU id spaces.** This PR's core insight is that `saveEndpoint.gpuIds` wants GPU **pool** ids (e.g. `ADA_24`), *not* the GPU **type** ids (e.g. `NVIDIA A40`) that `gpu list` and `--gpu-id` use. That's exactly why the PR added `ResolveServerlessGpuPoolID` — to translate the user's `--gpu-id` from type-id → pool-id before sending it.

**The gap.** When no `--gpu-id` is passed on the `--hub-id` path, the code falls back to the GPU ids baked into the hub release config — but assigns them **directly**, skipping that translation:
```go
if computeType == "GPU" && input.GpuIDs == "" && hubConfig.GpuIDs != "" {
    input.GpuIDs = hubConfig.GpuIDs   // <-- NOT run through ResolveServerlessGpuPoolID
}
```
So one source of GPU ids (`--gpu-id`) is translated, the other (hub config) is not.

**Why it may break.**
- *Before this PR:* `hubConfig.GpuIDs` was sent to REST's `gpuTypeIds` field, which expects **type** ids — so whatever the hub config stores worked there.
- *After this PR:* the same value is sent to `saveEndpoint.gpuIds`, which expects **pool** ids. If hub configs store type ids (which the old REST field strongly implies), `saveEndpoint` will reject them.

**Who hits it.** `runpodctl serverless create --hub-id <id>` **with no explicit `--gpu-id`** — i.e. relying on the hub's default GPU. That's a primary flow this PR is meant to support, so worth nailing down before merge.

**Ask / fix.** Please confirm what id-space hub configs actually store. If type ids, route this fallback through `ResolveServerlessGpuPoolID` as well — it returns pool ids unchanged, so it's safe whether the config holds type ids or pool ids.

### 🟠 Medium

**M1. Response selection drops `flashBootType` / `minCudaVersion` / `executionTimeoutMs`** — `internal/api/endpoints.go`, `CreateEndpointGQL` query. `--flash-boot=false` no longer echoes in `-o json` (old REST path set `endpoint.Flashboot` post-create) — a visible regression for scripting; cuda/timeout also never echo back. Add them to the selection set.

**M2. CPU always sends `flashBootType: FLASHBOOT`** — `create.go` (~L187). Flashboot block runs unconditionally (no `computeType` guard) and `--flash-boot` defaults `true`, so every default CPU create sends `FLASHBOOT`. e2e only verified `OFF` on GPU. Please test a default CPU create against live API; skip flashboot for CPU if the server rejects it.

**M3. `--scale-threshold 0` is unsendable** — `endpoints.go` (`ScalerValue ... json:"scalerValue,omitempty"`) + `create.go` (`if createScaleThreshold >= 0`). Sentinel is `-1`, so `0` is treated as set → `input.ScalerValue = 0` → then `omitempty` strips it. A threshold of 0 can't be sent.

**M4. `--env` silently ignored on the `--template-id` path** — `create.go`. `createEnvVars` is consumed only inside the `if createHubID != ""` block. `serverless create --template-id ... --env FOO=bar` is a silent no-op. Given the "no silent drops" goal and template-id being the common path, either reject `--env` with `--template-id` or wire it through.

**M5. `--min-cuda-version` (and `--gpu-count`) silently ignored on CPU** — `create.go`. Both only applied in the GPU branch. `--gpu-count` on CPU is nonsensical and correctly dropped (no warning); `--min-cuda-version` on CPU is a real silent ignore. Prefer an explicit error.

### 🟡 Low

- **L1.** No client-side `--name` length check (server requires ≥3). `--name ab` → raw server error. Generated fallback is fine. (`create.go` ~L305)
- **L2.** `--execution-timeout` has no range validation; `--idle-timeout` checks 1–3600. Asymmetric. (`create.go` ~L182)
- **L3.** Multi-region `--network-volume-ids` can't set per-volume `dataCenterId` — input struct has only `networkVolumeId`, though response struct + query carry `dataCenterId`. Missing-feature, not a bug. (`create.go` ~L201, `endpoints.go`)
- **L4.** GPU create does an extra `serverlessGpuPools` round-trip; a valid pool id like `ADA_24` still fails if that query errors (can't short-circuit). Consider a fast-path for obvious pool ids. (`gpu.go`)
- **L5.** `--instance-id` (CPU) sent verbatim, unvalidated → typo gives opaque graphql error (GPU path validates). `defaultCPUInstanceID="cpu3g-4-16"` hard-coded, drifts if server default changes. (`create.go` L18)
- **L6.** `math/rand/v2` → `math/rand` (v1) downgrade + `//nolint:gosec`. Harmless (names only, auto-seeded on Go 1.20+) but — why downgrade? (`create.go` L6)

TimPietruskyRunPod:
Thanks for the thorough review @justinwlin — addressed in `7b06c32` (CI green, re-verified live against a real account). Summary:

**🔴 H1 — hub `gpuIds` fallback untranslated** — Fixed. The hub-config fallback now goes through `ResolveServerlessGpuPoolID` too, and the resolver handles comma-separated lists (pool ids pass through unchanged, type ids translate, dedup). Verified live: `serverless create --hub-id <vllm>` with no `--gpu-id` now sends `gpuIds=ADA_80_PRO,AMPERE_80` and succeeds.

**🟠 M1 — response drops flashBootType/minCudaVersion/executionTimeoutMs** — Fixed. Added `computeType, flashBootType, minCudaVersion, executionTimeoutMs` (and already-present `instanceIds`, `networkVolumeIds`) to the `saveEndpoint` selection + `Endpoint` struct. `-o json` now echoes them — e.g. `--flash-boot=false` → `"flashBootType":"OFF"`, CPU → `"computeType":"CPU","instanceIds":["cpu3g-4-16"]`.

**🟠 M2 — CPU always sends `FLASHBOOT`** — Verified live: a default CPU create (`flashBootType: FLASHBOOT`) is accepted by the server and returns a working CPU endpoint, so no guard needed. Left as-is.

**🟠 M3 — `--scale-threshold 0` unsendable** — Turns out the server rejects it outright (`"Scaler value cannot be less than 1."`), so 0 is invalid, not just droppable. Now rejected client-side with `--scale-threshold must be at least 1` (caught before the API call). Valid values (e.g. `3`) confirmed sent as `scalerValue:3`.

**🟠 M4 — `--env` silently ignored on `--template-id`** — Fixed. Confirmed in the resolver that `saveEndpoint` ignores `input.env` when a `templateId` is referenced (env only applies via an inline template). Now prints `note: --env has no effect with --template-id (env is defined by the template); ignoring` to stderr instead of dropping silently.

**🟠 M5 — `--min-cuda-version`/`--gpu-count` ignored on CPU** — Now emit info notes to stderr (chose info over hard error so a stray flag doesn't block a create). Verified both notes fire on a CPU create.

**🟡 L1** — Fixed: `--name` < 3 chars rejected client-side.
**🟡 L2** — Left intentionally lenient: `--execution-timeout 0` appears valid server-side (no per-request limit), so no lower bound added.
**🟡 L3** — Not implementable client-side: the `NetworkVolumeIdsInput` GraphQL type only has `networkVolumeId`; the server derives `dataCenterId` from each volume. Current `[{networkVolumeId}]` shape is already complete.
**🟡 L4** — Improved: resolver falls back to passing the value through unchanged if the `serverlessGpuPools` query is unavailable, so a valid pool id still works during a transient pools-query failure.
**🟡 L5** — Added a light client-side sanity check on `--instance-id` (must look like a `cpu…-N-N` flavor) for a clearer error than the raw graphql one.
**🟡 L6** — Restored `math/rand/v2` and dropped the `//nolint:gosec`. The v1 downgrade was an accidental artifact of rebuilding the branch onto main during conflict resolution — not intentional.

Unit tests extended (validation table + full flag snapshot/restore). All paths re-verified live and cleaned up.

justinwlin (approved):

PR #286, opened 2026-06-11
fix: stderr for model output; GetModels empty case (on top of #276)

daveseddon-runpod:
## Summary

Follow-up fix-commit on top of #276 addressing two behavior-level regressions called out in review. Single commit, easy to cherry-pick — no feature changes, no scope creep.

**Recommendation**: same workflow as the previous follow-up (commit `2b7cf85`, which was cherry-picked onto the #276 branch) — please cherry-pick the single commit from this PR onto the `ce1579` branch as well, then close this PR. That keeps the #276 history linear and avoids competing/duplicate PR surface.

### Fixes

**Restore stderr for informational/error output** — CLAUDE.md says *"deprecation warnings go to stderr only (exec is the most common regression)."* PR #276 inadvertently moved these to stdout, which corrupts JSON output for agent consumers parsing stdout.

- `cmd/model/addModelToRepo.go` — the two "defaulting graphql timeout to N for model creation operations" notices go back to `os.Stderr`.
- `cmd/model/errors.go::handleModelRepoError` — uses `output.Error` (JSON to stderr) again instead of `fmt.Println` to stdout.

**Fix `GetModels` empty-result regression** — `api/model.go::GetModels` returns `[]*Model{}` again for an empty result, instead of `fmt.Errorf("data is nil: ...")`. Previously the CLI's `"no models found"` branch was unreachable; legitimate empty results surfaced as cobra errors.

### Tests

- `cmd/model/errors_test.go::TestHandleModelRepoError` (new, 4 rows): captures stdout + stderr via pipes, asserts stdout stays empty and the appropriate message lands on stderr across nil / not-implemented / feature-disabled / unrelated cases.
- `cmd/model/addModelToRepo_timeout_test.go::TestSetModelGraphQLTimeoutWithoutInheritedFlag`: extended with the same stdout-empty / stderr-non-empty check so the regression class can't recur silently for the timeout notice.

## Test plan
- [ ] \`go build ./...\` — passes locally
- [ ] \`go vet ./...\` — clean
- [ ] \`go test -count=1 ./...\` — all packages pass
- [ ] manual: \`runpodctl model add ...\` against the live API; capture stdout/stderr separately and confirm only result data lands on stdout

## Notes

Deliberately leaves the larger concerns from the review for separate PRs / brosenpod's call: the \`--output JSON default\` regression in model list/remove/add, the dropped \`DefaultGraphQLURL\` fallback + renamed env var, and the GQL fallback in \`serverless create\` silently dropping \`--flash-boot\`/\`--env\`/etc. Happy to follow up on any of those individually.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


justinwlin:
Closing as part of a backlog cleanup — this predates July and hasn't been reviewed. Not a rejection of the work: if it's still relevant, please reopen or rebase onto `main` and ping for review.

Issue #288, opened 2026-06-17
runpodctl serverless list fails to parse endpoints with attached network volumes


gabe-engineers:
## Summary

`runpodctl serverless list` fails to decode endpoints when the account has a Serverless endpoint attached to a network volume.

## Environment

- `runpodctl v2.0.0-beta.1`
- Reproduced on June 16, 2026

## Reproduction

1. Create a RunPod Serverless endpoint with an attached network volume.
2. Run:

```bash
runpodctl serverless list
```

## Expected

The CLI should return the list of Serverless endpoints.

## Actual

The CLI exits with a JSON unmarshal error:

```text
failed to parse response: json: cannot unmarshal string into Go struct field Endpoint.networkVolumeIds of type api.EndpointNetworkVolume
```

## Notes

- `runpodctl network-volume list` succeeds in the same environment.
- This looks like a schema mismatch around `networkVolumeIds` in the Serverless endpoint response.
- The failure appears to be client-side in `runpodctl`, though it may also indicate an API contract change/regression if the field shape no longer matches the client model.

gabe-engineers:
Additional observation from a live repro:

The raw REST response for `GET /v1/endpoints` is currently returning `networkVolumeIds` as a flat string array, for example:

```json
{
  "id": "svmgvz141knj88",
  "name": "quadra-bonsai",
  "networkVolumeId": "6fh4q4g8m5",
  "networkVolumeIds": ["6fh4q4g8m5"],
  "gpuIds": null,
  "gpuCount": 1,
  "workersMin": 0,
  "workersMax": 1
}
```

But `runpodctl` currently unmarshals that field as `[]EndpointNetworkVolume`:

```go
type Endpoint struct {
    NetworkVolumeIDs []EndpointNetworkVolume `json:"networkVolumeIds,omitempty"`
}

type EndpointNetworkVolume struct {
    NetworkVolumeID string `json:"networkVolumeId"`
    DataCenterID    string `json:"dataCenterId,omitempty"`
}
```

So a payload shaped like:

```json
"networkVolumeIds": ["6fh4q4g8m5"]
```

will reliably fail with:

```text
json: cannot unmarshal string into Go struct field Endpoint.networkVolumeIds of type api.EndpointNetworkVolume
```

This makes the issue look like a schema mismatch between the current REST response and the client model, not just a generic CLI parsing failure.

justinwlin:
[...]

PR #292, opened 2026-06-25
fix: parse serverless networkVolumeIds returned as id strings

TimPietruskyRunPod:
## problem

`runpodctl serverless get` and `serverless list` both fail for every endpoint that has a network volume attached:

```
failed to parse response: json: cannot unmarshal string into Go struct field Endpoint.networkVolumeIds of type api.EndpointNetworkVolume
```

the rest read endpoint returns `networkVolumeIds` as an array of bare id strings:

```json
"networkVolumeIds": ["x3s11yfqg9"]
```

but the `Endpoint` struct only accepted the graphql `saveEndpoint` **write** shape (`[{"networkVolumeId":"..."}]`). read and write shapes differ, so all serverless reads broke.

## fix

add a custom `UnmarshalJSON` on `EndpointNetworkVolume` that accepts either a bare id string (rest read) or an object (graphql write). no change to the write path.

## testing

- unit test covering both shapes (`TestEndpointNetworkVolumeIDsUnmarshal`)
- e2e against the live api: `serverless get` and `serverless list` now succeed on endpoints with attached network volumes (previously errored). read-only, no resources created.

### DIFF
--- a/internal/api/endpoints.go
+++ b/internal/api/endpoints.go
@@ -38,6 +38,25 @@ type EndpointNetworkVolume struct {
 	DataCenterID    string `json:"dataCenterId,omitempty"`
 }
 
+// UnmarshalJSON tolerates both shapes of networkVolumeIds: the rest read
+// endpoint returns bare id strings (["vol-1"]) while the graphql saveEndpoint
+// write path uses objects ([{"networkVolumeId":"vol-1"}]).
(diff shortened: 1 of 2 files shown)


justinwlin (approved):

gabe-engineers:
Closes https://github.com/runpod/runpodctl/issues/288

PR #294, opened 2026-06-26
fix: pod create template inheritance for containerDiskInGb, volumeInGb, and env overrides

jamesrahenry:
## Summary

Three silent failures when using `pod create --template-id`:

- **`--container-disk-in-gb` not inherited**: the CLI always sent the flag's default value of `20`, overriding the template's configured disk size with no error or warning.
- **`--volume-in-gb` not inherited**: unlike `containerDiskInGb`, the RunPod API does not propagate `volumeInGb` from the template when the field is absent — pods always came up with 0 (no volume), falling back to the 40 GB overlay instead of the dedicated network volume the template specified.
- **`--env` overrides silently dropped**: when `templateId` and `env` are both sent, the server merges template env on top of request env with template values winning for duplicate keys, making targeted per-pod overrides impossible without SSH + post-boot intervention.

## Changes

**`internal/api/graphql.go`**
- Add `omitempty` to `CreatePodGQLInput.ContainerDiskInGb` so the field is absent from the request when not explicitly set, allowing the server to inherit it from the template.

**`cmd/pod/create.go`**
- Zero out `createContainerDiskInGb` when `--container-disk-in-gb` is not explicitly provided, triggering the `omitempty` behaviour above.
- When `--template-id` is used, fetch the template before sending the create request. Use the fetch to: (1) inherit `volumeInGb` when `--volume-in-gb` is not explicitly provided, and (2) merge template env as the base with user-provided `--env` values winning for duplicate keys, then send the full merged set.

**`AGENTS.md`**
- Document the asymmetric server inheritance behaviour (`containerDiskInGb` is inherited, `volumeInGb` is not) and the env merge direction so future contributors don't inadvertently revert the workarounds.

## Test plan

All three fixes were verified end-to-end against the live API:

- [ ] `pod create --template-id <id>` (no `--container-disk-in-gb`) → pod `containerDiskInGb` matches template value, not 20
- [ ] `pod create --template-id <id>` (no `--volume-in-gb`) → pod `volumeInGb` matches template value, not 0
- [ ] `pod create --template-id <id> --env '{"KEY":"override"}'` → pod env contains override value for `KEY`; all other template env vars are present
- [ ] `pod create --template-id <id> --container-disk-in-gb 10` → explicit flag still respected
- [ ] `pod create --template-id <id> --volume-in-gb 50` → explicit flag still respected
- [ ] Unit tests pass: `go test ./...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


justinwlin:
Closing as part of a backlog cleanup — this predates July and hasn't been reviewed. Not a rejection of the work: if it's still relevant, please reopen or rebase onto `main` and ping for review.

PR #296, opened 2026-06-27
fix: fix serverless network volume decoding

### LINKED ISSUE #288: runpodctl serverless list fails to parse endpoints with attached network volumes

gabe-engineers:
## summary

- normalize `networkVolumeIds` decoding for serverless endpoints so rest string ids and graphql object entries both parse
- add regression coverage for `serverless list` responses containing attached network volumes

## root cause

`serverless list` decodes the rest `/endpoints` response into `Endpoint.NetworkVolumeIDs`, but the api can return `networkVolumeIds` as an array of strings while the client expected an array of objects. that made accounts with attached network volumes fail before output could be printed.

fixes #288

## validation

- `gocache=/tmp/go-build go test ./internal/api`
- `gocache=/tmp/go-build go build -o /tmp/runpodctl-bin/runpodctl .`
- `/tmp/runpodctl-bin/runpodctl serverless list` against the live api

## AI Disclaimer

Full disclaimer: I used AI for this and I'm not a Go pro, so I'm open to feedback, but looks reasonable to me.

### DIFF
--- a/internal/api/endpoints.go
+++ b/internal/api/endpoints.go
@@ -38,6 +38,24 @@ type EndpointNetworkVolume struct {
 	DataCenterID    string `json:"dataCenterId,omitempty"`
 }
 
+// EndpointNetworkVolume accepts both the REST list item shape (string) and the
+// GraphQL list item shape (object) for multi-region endpoint volumes.
+func (v *EndpointNetworkVolume) UnmarshalJSON(data []byte) error {
+	var id string
+	if err := json.Unmarshal(data, &id); err == nil {
+		*v = EndpointNetworkVolume{NetworkVolumeID: id}
+		return nil
+	}
+
+	type endpointNetworkVolume EndpointNetworkVolume
+	var volume endpointNetworkVolume
+	if err := json.Unmarshal(data, &volume); err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


gabe-engineers:
Just realized https://github.com/runpod/runpodctl/pull/292 would solve the issue, will close this then

PR #299, opened 2026-07-08
fix: STO-333: Centralize api env/config lookup

brosenpod:
# STO-333: Centralize api env/config lookup

Add configenv helpers for api key, rest url, and graphql url lookup
Route rest, graphql, doctor, and legacy query paths through them

Previously env vars were inconsistently used as overrides across the project. This makes them consistently override other options.


brosenpod:
@cusor review

PR #302, opened 2026-07-10
STO333, STO334, STO335

brosenpod:
Fixes issues with env vars not alway overriding config values, and centralizes how we track config values
Adds a progress bar to Model Repo uploads and adds an option to wait until the model is hashed before returning
Allows removing a specific Model version from Model Repo

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -172,6 +172,7 @@ type ModelVersionStatusMutationResult struct {
 type AddModelToRepoInput struct {
 	Owner               string                 `json:"owner,omitempty"`
 	Name                string                 `json:"name"`
+	Provider            string                 `json:"provider,omitempty"`
 	CredentialType      string                 `json:"credentialType,omitempty"`
 	CredentialReference string                 `json:"credentialReference,omitempty"`
 	ModelStatus         string                 `json:"modelStatus,omitempty"`
@@ -234,6 +235,7 @@ func AddModelToRepo(input *AddModelToRepoInput) (*Model, error) {
 	}
 
 	addString("owner", input.Owner)
+	addString("provider", input.Provider)
 	addString("credentialType", input.CredentialType)
 	addString("credentialReference", input.CredentialReference)
 	addString("modelStatus", input.ModelStatus)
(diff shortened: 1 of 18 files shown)


murat-runpod:
@brosenpod any chance you could drop some screenshots or test command of the progess bar?

brosenpod:
@murat-runpod Of course! I should have had that in the description. Looks like this:
```
% ./bin/runpodctl model add \
    --name "$NAME" \
    --model-path ./foo \
    --wait-for-hash \
    --create-upload

uploading model 100% |████████████████████| (43/43 B,  4 B/s)
model size: 43 bytes
waiting for model to be hashed.................................
model is ready to deploy, your model url is: "https://local/user_2srUX49gvgYFwRMNH5N2JbmqxVi/Super-Cool-Test-1783718166:9e6aced02739cf261816c64a40098cb6fd5e57bd2320c2d9927ab36d93c984f4"
{
  "model": {
    "id": "user_2srUX49gvgYFwRMNH5N2JbmqxVi/Super-Cool-Test-1783718166",
    "name": "Super-Cool-Test-1783718166",
    "owner": "user_2srUX49gvgYFwRMNH5N2JbmqxVi"
  }
}
```

murat-runpod (approved):
This stood out from Claude review - up to you if you think it's relevant to end-users:

> Medium — --wait-for-hash output hides its own deliverable
> 
> - In non-verbose mode the resolved model URL and hash are written only to stderr (printModelReadyURL) and are omitted from the stdout JSON — compactModelAddOutput carries just id/name/owner (cmd/model/addModelToRepo.go:290-309, 519-524). Since the default output is JSON for agent consumption, an agent that runs --wait-for-hash specifically to obtain a deployable reference cannot read it from stdout without also passing --verbose. Consider adding modelUrl/modelHash to the compact struct so the wait's result is machine-readable by default.
> 

Otherwise LGTM!

brosenpod:
@murat-runpod Yeah, that is known. The `--wait-for-hash` flow is made for human consumption. This project currently treats every user as an agent and no humans, which we are tracking in a different ticket for perhaps changing. An agent can figure out that it shouldn't use `--wait-for-hash` and that it needs to poll on its own.

PR #303, opened 2026-07-13
fix: verify self-update archive checksums

k-nox:
## Summary

- Verifies self-update release archives against GoReleaser's versioned SHA-256 checksum asset before extraction or replacement.
- Applies the same fail-closed checksum verification to `install.sh` and removes temporary download artifacts after both successful and failed installations.
- Adds Go and shell coverage for checksum parsing, mismatch handling, verification ordering, and installer cleanup.

## Testing

- `go test ./... -count=1`
- `go build ./...`
- `bash -n install.sh`
- `bash -n install_test.sh`
- `bash install_test.sh`
- Built a disposable binary in a temporary directory and ran `update` against live GitHub release assets without replacing a production installation.
- Mutation-tested the installer cleanup regression coverage by removing the cleanup trap and confirming the test failed for the expected leftover directory.

## Security considerations

- Verification occurs before archive extraction and binary replacement.
- Missing checksum assets, missing archive entries, malformed digests, mismatches, and unavailable checksum tooling fail closed.
- This protects against accidental corruption or replacement of the archive alone. A compromise that replaces both the release archive and its checksum asset remains out of scope and would require signed release metadata.

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -4,6 +4,8 @@ import (
 	"archive/tar"
 	"archive/zip"
 	"compress/gzip"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


cmd/update.go
```diff
@@ -213,13 +279,23 @@ var updateCmd = &cobra.Command{
...
 		fmt.Printf("downloading runpodctl %s\n", latestVersion)
-		file, err := DownloadFile(downloadLink, archivePath)
+		file, err := DownloadFile(downloadAsset.Url, archivePath)
 		if err != nil {
 			fmt.Println("error fetching the latest version of runpodctl:", err)
 			return
 		}
 		file.Close()
 
+		checksumText, err := DownloadBytes(checksumAsset.Url)
```
brosenpod, line 285: Checksum is not authenticated: The updater downloads the archive and its checksum from the same mutable GitHub release. An attacker able to replace the archive can replace the checksum asset too, so verification still succeeds. The installer has the same flaw at (install.sh:198). Use a signed provenance/checksum verified against an embedded public key, or pin a trusted digest from an independently authenticated channel.

justinwlin (approved):
Rebased onto main and resolved conflicts with #306 (Run -> RunE); converted the three new checksum failure paths to wrapped `return fmt.Errorf(...)` so a failed verification exits nonzero.

Verified against live v2.9.0 release assets:
- checksum file format is `<hash>  <name>`, and every archive name the code requests is present
- end-to-end `update` from a binary built at version 1.0.0: downloaded, verified, replaced, exit 0
- tampered archive (1 byte appended) correctly rejected by both the Go and install.sh paths
- `go vet`, `go build`, `go test ./...`, `govulncheck ./...` all clean; `bash install_test.sh` exit 0

Two follow-ups, neither introduced here:
1. Windows self-update is broken independently of this PR — the `defer os.Remove(extractedPath)` in cmd/update.go removes the `.new` file before the user can run the printed `move` command. Same in main.
2. This verifies integrity, not authenticity: the checksum asset comes from the same unauthenticated release as the archive, so it covers corruption rather than release compromise. Closing the "unsigned self-update" concern fully needs signed release metadata (cosign/minisign or GitHub attestations).

PR #304, opened 2026-07-14
fix: STO-356: Use hash in model repo GraphQL queries

brosenpod:
Remove the obsolete ModelVersion.versionHash selection and client field. Use hash as the sole model-version identifier for listing, lookup, status updates, and upload hash polling. Update unit tests for the current schema.

This resolves the issue with `--wait-for-hash` using an incorrect GQL query, causing a failure.

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -38,13 +38,12 @@ type ModelUser struct {
 
 // ModelVersion represents a specific version of a model stored in the repository.
 type ModelVersion struct {
-	UUID        string                 `json:"uuid,omitempty"`
-	Hash        string                 `json:"hash,omitempty"`
-	VersionHash string                 `json:"versionHash,omitempty"`
-	Status      string                 `json:"status,omitempty"`
-	CreatedAt   string                 `json:"createdAt,omitempty"`
-	UpdatedAt   string                 `json:"updatedAt,omitempty"`
-	Metadata    map[string]interface{} `json:"metadata,omitempty"`
+	UUID      string                 `json:"uuid,omitempty"`
+	Hash      string                 `json:"hash,omitempty"`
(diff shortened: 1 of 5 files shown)


daveseddon-runpod:
## Code Review — Approve ✅

**Overview:** Removes the obsolete `ModelVersion.versionHash` field and its GraphQL selections. Since the backend schema no longer exposes `versionHash`, requesting it broke the query — which is the root cause of the `--wait-for-hash` failure. Standardizes on `hash` as the sole version identifier across listing, lookup, status updates, and upload hash polling.

### Correctness
- Root-cause fix is sound: the three query bodies (`GetModels`, `GetModel`, `newUpdateModelVersionStatusInput`) no longer select the invalid field.
- Struct + all consumers cleaned up consistently (`uploadedModelVersionHash`, `removeModelVersion`, `findModelVersion`).
- Verified no stray references remain — grep shows only unrelated function names and the new negative-assertion test.
- `go build ./...` clean; `go test ./api/... ./cmd/model/...` green.

### Tests
- `TestGetModelsRequestsVersionIdentifiers` now asserts the query requests `hash` **and** must *not* contain `versionHash` — good regression guard.
- Legacy-fallback test cases dropped appropriately, since the field can no longer be populated.

### Notes
- The removed `VersionHash` fallback was already unreachable (server stopped returning the field), so this is correct cleanup rather than a behavior change — worth a one-line mention in the description.
- Per project policy, suggest a quick live `--wait-for-hash` run against the real API to close out the original bug report before merging; unit tests cover the query shape but not the live round-trip.

**Risk: Low** — purely subtractive, no new flags/paths/output changes.

daveseddon-runpod (approved):

PR #305, opened 2026-07-14
fix: STO-354: Preserve Hub deployment constraints on endpoint create

brosenpod:
# Preserve Hub deployment constraints on endpoint create

Apply Hub `runsOn`, GPU count, and CUDA requirements when creating serverless endpoints from a Hub release, while preserving explicit CLI overrides. Fail clearly on invalid release config and add focused coverage for config parsing and precedence.

### DIFF
--- a/cmd/serverless/create.go
+++ b/cmd/serverless/create.go
@@ -23,6 +23,7 @@ var createCmd = &cobra.Command{
 
 requires either --template-id or --hub-id.
 --hub-id accepts both SERVERLESS and POD hub listings.
+hub deployment constraints are applied unless explicitly overridden by cli flags.
 
 examples:
   # create from a template
@@ -69,6 +70,16 @@ var (
 	createNetworkVolumeIDs string
 	createModelReferences  []string
 )
+
+type serverlessCreateClient interface {
+	GetListing(string) (*api.Listing, error)
+	ResolveServerlessGpuPoolID(string) (string, error)
+	CreateEndpointGQL(*api.EndpointCreateGQLInput) (*api.Endpoint, error)
+}
+
+var newServerlessCreateClient = func() (serverlessCreateClient, error) {
+	return api.NewClient()
+}
 
 func init() {
 	createCmd.Flags().StringVar(&createName, "name", "", "endpoint name")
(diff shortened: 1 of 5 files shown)


daveseddon-runpod:
## Code Review — Approve with questions

**Overview:** Applies Hub release deployment constraints (`runsOn`, `gpuCount`, `allowedCudaVersions`) on serverless create while letting explicit CLI flags win. Refactors validation into shared helpers (`validateComputeFlags`, `flagChanged`, `hubGPUCount`, `hubMinCudaVersion`), moves the listing fetch/parse earlier so `computeType` resolves before input build, and removes the previous double-parse of `release.Config`. Build clean; `go test ./cmd/serverless/... ./internal/api/...` green.

### Correctness
- Precedence logic is right: Hub value used only when the flag wasn't changed and the Hub value is >0; `--min-cuda-version`/`--gpu-count` override correctly (verified against the unit tests).
- `computeType` resolution ordering is correct — Hub `runsOn` applied (only when `--compute-type` unset) before the CPU/GPU branch and before validation.
- No double `GetListing`/double parse — consolidated into the early block; later hub block reuses the resolved vars. Good cleanup.
- Strict config parse is a deliberate behavior change (malformed `release.Config` now hard-fails instead of falling back). Matches intent, but worth a release-note callout.

### Issues / questions
- **`runsOn` value space (please verify):** L172-173 rejects any non-`GPU`/`CPU` `runsOn`. The description says `--hub-id` accepts SERVERLESS **and POD** listings — confirm a valid POD release can't carry a `runsOn` outside `{GPU, CPU, \"\"}`, else this starts rejecting previously-working POD deploys. (Empty is safely guarded.)
- **Confusing error when constraint is Hub-sourced:** if Hub `runsOn=CPU` and the user passes `--gpu-id` without setting `--compute-type`, they get `--gpu-id must be empty when --compute-type is CPU` despite never setting `--compute-type`. Consider naming the Hub release as the source.
- **Redundant validation (minor):** `validateComputeFlags` runs at L124-128 and again unconditionally at L175. The first is only a pre-network fast-fail; a comment (or dropping it) would clarify intent.
- **Style nit:** `validateComputeFlags(computeType, gpuTypeID)` mixes params with package globals (`createInstanceID`, `createModelReferences`); passing all inputs would be more consistent/testable.

### Test coverage
- Good unit coverage for the pure helpers and config JSON parsing.
- **Gap:** no test exercises the integration-level fix inside `runCreate` — `runsOn` overriding `computeType`, the validate-skip-when-hub logic, or `MinCudaVersion`/`GpuCount` application. Requires mocking the client, but that's where a regression would land.

### Project conventions
- **E2E is required here** — this changes the live `saveEndpoint` create path from a Hub release. Please build and run `serverless create --hub-id …` against the live API (GPU and, if possible, a CPU `runsOn` release), confirm constraints apply and overrides are honored, and delete the endpoint immediately.
- **Skill sync:** `--hub-id` behavior changed — update the runpodctl skill.
- Docs regenerated and lowercase-output convention respected. 👍

**Risk: Medium** — sound, well-factored logic, but touches the money-spending create path and adds two behavior changes (strict config parse; `runsOn` validation) that could reject previously-accepted inputs. Clear the `runsOn` value-space question and run E2E before merge.

brosenpod:
@daveseddon-runpod 

`runsOn` must be either "GPU" or "CPU", according to spec:
https://docs.runpod.io/hub/publishing-guide#runpod-configuration

Updated:
* Hub-derived CPU conflicts now identify the Hub listing as the source.
* Documented the intentional pre-network validation.
* Made `validateComputeFlags` fully parameterized.
* Added mocked `runCreate` coverage for Hub `runsOn` override, deferred validation, GPU count, and CUDA minimum application.

daveseddon-runpod:
## Re-review — LGTM ✅

Thanks @brosenpod, all four points are addressed and the build/tests are green on the updated branch.

- **`runsOn` value space** — confirmed against the [publishing spec](https://docs.runpod.io/hub/publishing-guide#runpod-configuration) that `runsOn` is constrained to `GPU`/`CPU`, so rejecting other values is correct. Question resolved.
- **Hub-sourced constraint error** — the `hub listing %q requires %s: %w` wrapping (gated on the compute type being Hub-derived) makes the failure much clearer. 👍
- **`validateComputeFlags`** — now fully parameterized (`computeType, gpuTypeID, instanceID, modelReferences`); no more mixing params with package globals.
- **Integration coverage** — `TestRunCreate_HubDeploymentConstraints` via the injectable `serverlessCreateClient` mock now exercises the actual `runCreate` fix: `runsOn` override → CPU instance, GPU count + CUDA-min application, and the deferred Hub-derived validation (including asserting the Hub lookup ran before the compute check). This closes the gap I flagged.
- The pre-network validation comment clarifies the intentional double-check.

Remaining before merge (unchanged from prior review, both outside the diff):
- **E2E on the live create path** — please still run `serverless create --hub-id …` against a real GPU and a CPU-`runsOn` release, confirm constraints apply / CLI overrides win, and delete the endpoints immediately.
- **Skill sync** — update the runpodctl skill for the new `--hub-id` constraint-inheritance behavior.

Approving from a code standpoint.

daveseddon-runpod (approved):

PR #306, opened 2026-07-15
feat: agent-friendly cli output (CON-683, CON-684, CON-685)

justinwlin:
## Summary

Agent-friendly output improvements to the runpodctl v2 surface, covering three linked tickets from the CON_2026 Q3 Agents Emergency Triage rock. All three are "make the CLI legible to an agent driving it."

- **CON-683 — clean CLI error output**
- **CON-684 — invoke URLs in serverless output**
- **CON-685 — pricing + per-datacenter availability in `gpu list`**

## CON-683: clean CLI error output

Before, a runtime failure printed the error JSON, then the full flag/usage dump, then the error again — and API errors were double-encoded (`{"error":"api error: {\"error\":\"pod not found\"} (status 404)"}`), so agents misread runtime failures as flag mistakes.

- Set `SilenceUsage` (`SilenceErrors` was already set). Usage is now re-printed **only** for genuine usage errors (bad flags, wrong arg count, unknown command) via a `usageError` type + `SetFlagErrorFunc`.
- `Execute` is now the single error sink: one flat JSON error object on stderr with a stable `code` field (and HTTP `status` when available). Removed the redundant per-command `output.Error(err)` calls (88 sites).
- The REST client unwraps nested API error JSON: `APIError` is now a real `error` type carrying the unwrapped message + code + status.

```
# runtime error (unwrapped, stable code, no usage dump, printed once)
$ runpodctl serverless get does-not-exist
{"error":"failed to get endpoint: endpoint not found","code":"not_found","status":404}

# usage error (still shows usage, tagged usage_error)
$ runpodctl serverless get
{"error":"accepts 1 arg(s), received 0","code":"usage_error"}
Usage:
  runpodctl serverless get <endpoint-id> [flags]
  ...
```

## CON-684: invoke URLs in serverless output

`serverless create` / `get` / `list` now include a `urls` object (`run`, `runsync`, `health`) computed client-side from the endpoint id, so an agent that just deployed an endpoint can call it without a UI round-trip.

```
$ runpodctl serverless get jlbr0gof5j76wh
{ ..., "urls": {
  "run":     "https://api.runpod.ai/v2/jlbr0gof5j76wh/run",
  "runsync": "https://api.runpod.ai/v2/jlbr0gof5j76wh/runsync",
  "health":  "https://api.runpod.ai/v2/jlbr0gof5j76wh/health"
}}
```

## CON-685: pricing + per-datacenter availability in `gpu list`

Added secure/community on-demand `$/hr` (`securePricePerHr`, `communityPricePerHr`) and a per-datacenter availability breakdown alongside the existing best-overall `stockStatus`, so agents can cost-optimize and pick a DC that will actually schedule.

```
$ runpodctl gpu list
[{ "gpuId": "NVIDIA A100 80GB PCIe", "securePricePerHr": 1.39, "communityPricePerHr": 1.19,
   "stockStatus": "Low",
   "dataCenterAvailability": [{"dataCenterId":"CA-MTL-3","stockStatus":"Low"}, ...] }, ...]
```

## Testing

- `go build ./...`, `go vet ./...`, `go test ./...` — 237 pass.
- Added unit tests: `parseAPIError` (unwrap/code/status), `output.Error` with code/status + nil no-op, usage-error classification + `SilenceUsage`/`SilenceErrors`, `invokeURLs` + get/list URL population, `ListGpuTypes` pricing + per-DC aggregation.
- **e2e against the live API** (`RUNPOD_API_KEY`): `gpu list` (pricing + per-DC), runtime/usage/unknown-command/unknown-flag error shapes, and full `serverless create` → `get` → `list` → `delete` cycle (created a temporary serverless template + endpoint with `workers-min 0`, verified `urls`, then deleted both — no lingering resources).

## Follow-up

- The runpodctl agent skill (runpod/skills) should get a small refresh for the new `gpu list` fields, `urls`, and error shape. Not in this PR.

Kept as a draft for your review.

---

## ⚠️ Behavior changes

Both are bug fixes, but they can break a script that relied on the bug:

1. **`model` and `update` now exit non-zero on failure.** They previously printed the error and exited **0**, so `runpodctl model add … && next-step` ran `next-step` after a real failure.
2. **`api/query.go` no longer prints an error to stdout.** It printed `API key not found` to *stdout* while also returning the error, so anything piping `model` output into a JSON parser got that line mixed in.

3. **`send` and `receive` now exit non-zero on failure.** `send` had a path that produced *no output at all* and exited 0, so `runpodctl send x && echo ok` printed `ok` on a silent failure. `receive` was on `Run:` with five `Fatalf` sites.
4. **`template search` / `hub search` emit `[]` instead of prose when nothing matches.** They printed `no templates found matching "x"` to **stdout**, which broke callers piping into a JSON parser. Anything grepping stdout for that sentence needs updating.

`update` also surfaces a previously fire-and-forget `mv` failure, so a botched install is no longer silent.

## Changes after review

Two audit passes over this branch found more instances of the bug class this PR exists to fix. All verified with build/vet/tests green and, where possible, against the live API.

**Errors that bypassed the sink**
- `cmd/ssh.go` and `cmd/transfer/transfer.go` (`send`) printed their own error *and* returned it — double-print, or in send's case a bare `return` producing no output and exit 0. Both used hand-rolled prints rather than `output.Error`, which is why the original sweep missed them.
- `required flag(s) … not set` had no code and no usage text: Cobra's `ValidateRequiredFlags` runs after flag parsing and never reaches `SetFlagErrorFunc`. 20 sites affected.
- `CreatePod` returned a raw untyped graphql error — the last of 19 such sites.

**The `code` promise now actually holds.** ~229 hand-written `fmt.Errorf` sites and every transport failure emitted a bare `{"error":"..."}`. Uncoded errors are now classified in the sink: `network_error` (unreachable API, detected via `errors.As` on `*url.Error`/`*net.OpError`/`*net.DNSError`, excluding url *parse* failures — those are `cli_error`, since `network_error` is the one code meaning "retry"), `no_credentials` (no API key), `cli_error` (everything else local). A typed error's own code always wins.

**A latent wrong-answer bug in CON-685.** `betterStock` looked statuses up in a map and Go returns 0 for a missing key, so any unrecognized value tied with "no stock" and lost to `Low` — a new API stock level or a casing change would have made the top-level `stockStatus` under-report an available GPU. `stockRank` is now case-insensitive and ranks unknown above absent; `TestStockRank` pins the known set.

**Docs.** README gains an error-format section (shape, the codes, and why agents must branch on `code` rather than `status`, since graphql not-founds carry no status) and an env-var table — none of this was documented anywhere. `docs/` regenerated.

**Also:** `pod create`'s global-networking error now carries the `--data-center-ids` hint in the message; it previously reached the user only via the usage dump this PR removes (caught by the `-tags e2e` suite, which CI does not run).

**New tests:** `TestCodeForStatus` (pins the vocabulary — four codes had never been asserted), `TestStockRank`, `TestErrorFallbackCode`, `TestFallbackCode_ParseErrorIsNotNetwork`, `TestErrorTypedCodeWinsOverFallback`, `TestModelRepoError`, `TestModelCommandsUseRunE`, invoke-URL edge cases.

**Not done here:** unifying the two "no stock" sentinels (per-DC says `"none"`, top-level omits the key) — that changes the shape of a pre-existing field for about half of `gpu list`, so it is a contract decision, not a bug fix. And `model add`'s remaining `cobra.CheckErr` sites still emit plaintext; a stacked PR covers those.

**Follow-ups:** CON-752 (agent skill needs the new codes), CON-804, CON-805.

## Round 2 — after @LarveyOfficial's review

His two blockers, plus the rest of his list. Full replies in the thread.

- **Every missing-key path now reports `no_credentials`.** `NewGraphQLClient` returned an untyped error, so `ssh *`, `pod create`, `pod get` and `template create/update` said `cli_error` while `pod list` said `no_credentials`.
- **`api.ErrNoCredentials` is re-exported from the public `api` package** as the same value. Returning an `internal/` sentinel meant external importers couldn't `errors.Is` against it.
- **README no longer claims the JSON shape is universal.** It now carries a known-gaps table: the legacy pod commands and `exec` still print plaintext, and `project` prints to stdout and exits 0 (CON-816).
- **`send` finished and `receive` converted** — four `Fatalf` exits remained in `send`, and `receive` was untouched.
- **`stockRank` can no longer report a GPU as available with no stock.** Unknown non-empty statuses rank above absent, so a literal `none`/`unavailable` would have won the top-level `stockStatus` and set `available: true`. Those literals now rank 0, and `available` derives from `stockRank` via a shared `hasStock()`.
- **`parseAPIError` can no longer emit a nested JSON blob** — it reads `detail` as well, and falls back to the status line rather than dumping a body it can't read a message out of.
- **`modelRepoError` deleted** — after the refactor its only remaining effect was replacing a wrapped error with a bare sentinel.
- **Prices are `*float64` with an explicit `null`** when a cloud type doesn't offer the GPU. With `omitempty` on a `float64`, "not offered" and a real `0` were identical; six GPUs in today's `gpu list` hit that case.

### Two things deliberately not done

- **The GPU price tags were NOT renamed** (his 9c). I tried it: those tags are the *decode* contract for the GraphQL response (`securePrice`), so adding a `PerHr` suffix makes `json.Unmarshal` miss the fields and silently report every price as **0**. Backed out, verified against live data, and left a `DO NOT rename` comment. Nothing outside `internal/api` marshals GPU data, so users only ever see the `PerHr` names.
- **The exit-code fixes were not split into their own PR** (his 10). The repo squash-merges, so this lands as one commit either way; a targeted forward-fix later is cheaper than splitting and re-reviewing 20 commits.

### The e2e suite is green end to end

It stays out of CI deliberately — ~7 tests provision real pods and endpoints, so a per-PR run costs money and a test dying before cleanup leaks a pod. But it's now trustworthy: the one test failing on `main` pinned a Hub template that has been delisted, and fixing it surfaced the prose-on-stdout bug above. Replaced with two tests that assert the contract rather than a third party's continued existence. **90/90 pass.**

If anyone wires it up later, split it by cost: a read-only tag per-PR, a resource-creating tag nightly. The `pod create --global-networking` regression this PR caused would have been caught by the free half.


justinwlin:
## Verification

Verified with unit tests + live e2e against the real API. All checks green.

**Automated:** `go build ./...`, `go vet ./...`, `go test ./...` → **all tests pass**. New tests cover `parseAPIError` (unwrap/code/status), `parseGraphQLHTTPError`, `output.Error` code/status + nil no-op, `asUsageError` classification incl. typed-error short-circuit, `invokeURLs` + get/list population, and `ListGpuTypes` pricing + per-DC aggregation (including the empty-stock → `"none"` case).

**Live e2e** (real `RUNPOD_API_KEY`), captured from the built binary:

### CON-685 — `gpu list` pricing + per-DC availability
```json
{
  "gpuId": "NVIDIA A100 80GB PCIe",
  "displayName": "A100 PCIe",
  "memoryInGb": 80,
  "secureCloud": true,
  "communityCloud": true,
  "securePricePerHr": 1.39,
  "communityPricePerHr": 1.19,
  "stockStatus": "Low",
  "available": true,
  "dataCenterAvailability": [
    { "dataCenterId": "CA-MTL-3", "stockStatus": "Low" },
    { "dataCenterId": "EU-RO-1",  "stockStatus": "Low" },
    { "dataCenterId": "US-KS-2",  "stockStatus": "Low" },
    { "dataCenterId": "US-MO-1",  "stockStatus": "none" }
  ]
}
```
Every datacenter the GPU appears in is listed; a datacenter with no reported stock is shown explicitly as `"none"` (not hidden, not an empty string) so an agent can tell "offered here, currently no stock" from a missing entry.

### CON-684 — invoke URLs on serverless `get` (live endpoint)
```json
{
  "id": "jlbr0gof5j76wh",
  "name": "live-whisper-flash",
  "urls": {
    "run":     "https://api.runpod.ai/v2/jlbr0gof5j76wh/run",
    "runsync": "https://api.runpod.ai/v2/jlbr0gof5j76wh/runsync",
    "health":  "https://api.runpod.ai/v2/jlbr0gof5j76wh/health"
  }
}
```
`create` and `list` were verified the same way — I stood up a temporary serverless template + endpoint (`--workers-min 0`, no cost), confirmed `urls` on create/get/list, then deleted both (no lingering resources).

### CON-683 — clean error output
```
# runtime error: unwrapped, stable code + status, printed once, no usage dump
$ runpodctl serverless get nope-nope
{"error":"failed to get endpoint: endpoint not found","code":"not_found","status":404}

# usage errors: JSON with code + usage text
$ runpodctl serverless get
{"error":"accepts 1 arg(s), received 0","code":"usage_error"}
  ...usage...

$ runpodctl gpu list --bogus
{"error":"unknown flag: --bogus","code":"usage_error"}
  ...usage...
```

Follow-up to sync the runpodctl agent skill after merge: CON-752.

LarveyOfficial (commented):
## Overview

Three linked agent-legibility improvements to runpodctl v2: (683) a single flat JSON error sink with stable `code`/`status` fields, `SilenceUsage`, unwrapped API/GraphQL errors, and removal of ~88 redundant per-command `output.Error` calls; (684) client-computed invoke `urls` on serverless `create`/`get`/`list`; (685) on-demand pricing + per-datacenter availability in `gpu list`.

## Strengths

- **The error-handling redesign is the right architecture.** Making `APIError`/`GraphQLError` real error types with `ErrorCode()`/`HTTPStatus()`, then having `output.Error` pull those out via `errors.As` through the wrap chain, means a wrapped `fmt.Errorf("failed to X: %w", apiErr)` still surfaces the correct code/status. Collapsing 88 call sites into a single `Execute` sink is a genuine consistency win and removes the double-print/usage-dump bug.
- **`SilenceUsage` + typed `usageError` split is correct.** Usage text now shows only for real usage mistakes, and the `SetFlagErrorFunc` route + `nil` no-op guard in `output.Error` are good touches.
- **Good, targeted tests** — `parseAPIError` (unwrap/code/status), the "must not be double-encoded" assertions, usage-vs-runtime classification including the typed-error bail, and the per-DC "none" vs missing distinction.

## Concerns / suggestions

1. **`asUsageError` string-prefix fallback is the main fragility.** After the (good) typed `APIError`/`GraphQLError` bail-out, any *plain* `fmt.Errorf` runtime error whose message happens to start with `"invalid argument"`, `"accepts "`, `"requires at least"`, etc. would be misclassified as a usage error and dump usage text. Today's plain runtime errors don't collide (e.g. `pod/list.go`'s `"invalid --created-after format…"` starts with `"invalid "`, not `"invalid argument"`), but this couples correctness to Cobra's internal message strings *and* to every command author avoiding those prefixes. `TestAsUsageError` guards the Cobra strings, which helps — consider a short comment warning command authors off those leading words, and note it may need revisiting on a Cobra upgrade.

2. **Hardcoded invoke base `https://api.runpod.ai/v2`** in `invokeURLs`. Correct as long as the serverless invoke domain is always prod-global and independent of the control-plane API URL. If someone points runpodctl at a dev/staging API, the emitted `urls` would still target prod — worth a one-line confirmation that invoke is always this host regardless of `apiUrl`.

3. **GraphQL errors collapse to a single `graphql_error` code with no status** (200-with-`errors` body). Acceptable and documented, but it means an agent can't branch a GraphQL "not found" the way it can a REST 404. Fine as a known tradeoff — flagging only so it's a conscious one.

## Nits

- The `APIError` field rename `Error` → `Message` (method vs field) is contained to `internal/` with no external importers and `go build`/`go vet` pass, so no compatibility risk — just the kind of change worth a line in the description.
- PR is still marked **draft**; scope (51 files, 3 tickets) is large but cohesive. No objection to the bundling given they're one theme.

Solid, well-tested work with the right architecture. Only #1 (string-prefix fragility) is worth a follow-up thought; nothing blocking.

justinwlin:
Rebased onto main and expanded after review. Short version:

**@LarveyOfficial your three items:** (1) usage-prefix fragility — documented, plus verified no error message in the repo actually starts with those prefixes. (2) hardcoded invoke base — made overridable via `RUNPOD_INVOKE_URL` instead of just confirming it. (3) graphql not-found — now returns typed `not_found` so agents can branch on it; the general "one code, no status" point stands by design.

**Then two audit passes found more of the same bug class, all fixed:**
- `ssh info` and `runpodctl send` printed their own error *and* returned it → double-print / silent exit 0
- `required flag(s) …` had no code and no usage text
- ~229 validation errors and all transport failures had no `code` at all — now `cli_error` / `network_error` / `no_credentials`, assigned in the sink
- `betterStock` scored unknown stock levels as 0, tying with "no stock", so top-level `stockStatus` could under-report an available GPU

**⚠️ Two behavior changes.** Both were bugs rather than intended contracts, so I think shipping them is fine — but they are observable, so worth stating plainly.

*1. `model` / `update` exited 0 on failure; they now exit non-zero.*

```bash
runpodctl model add --name foo … && echo uploaded
# before: prints the error, then prints "uploaded"   ← && believed it succeeded
# after:  prints the error, stops
```

*2. `api/query.go` printed an error to stdout; it no longer does.*

```bash
runpodctl model list | jq '.[].name'
# before: stdout = "API key not found" + json   ← jq chokes
# after:  stdout = json only; error on stderr as
#         {"error":"api key not found…","code":"no_credentials"}
```

Why this is fine despite changing observable behavior: an exit 0 on failure isn't an interface anyone could have deliberately depended on, and an error on stdout contradicts the documented contract (data on stdout, errors on stderr) that the rest of the CLI already follows. In both cases the new output is strictly easier to consume — one json object, on the right stream, with a stable `code`. Anyone affected was relying on a broken path, and their pipeline was already failing silently; this makes it visible. Documented in the README error-format section, and cc @brosenpod since `cmd/model/*` is Model Repo surface and he'd field any report.

**Verified:** build/vet/unit tests green; the `-tags e2e` suite green apart from one pre-existing failure on main; live e2e for all three tickets (`gpu list` pricing + per-DC, `serverless create/get/list/delete` with `urls`, `GET /v2/<id>/health` → 200, every error class), resources cleaned up.

Follow-ups filed: CON-804 (`serverless update --workers-max 0` silently no-ops — `omitempty` drops the zero), CON-805 (`pod create` lacks `--model-reference`), CON-752 (skill needs the new codes). A stacked PR is coming for the remaining `model` plaintext paths.

LarveyOfficial:
## Claude Review

Solid work. The single-sink refactor is genuinely clean: 90 `output.Error` calls removed, 2 added, both in `Execute`, and I checked every removal site — each one is immediately followed by a `return` of the same (or wrapped) error, so nothing got swallowed. The `url.Error` / `Op != "parse"` gate and the typed-error short-circuit in `asUsageError` are both subtle catches that most people would have missed. `go build` and `go vet` are clean on the branch.

Findings below, roughly highest-value first.

### 1. `NewGraphQLClient` still returns an untyped "api key not found" — `code: cli_error`, not `no_credentials`

[`[internal/api/graphql.go:39](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/internal/api/graphql.go#L39)`](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/internal/api/graphql.go#L39)

```go
return nil, fmt.Errorf("api key not found. run 'runpodctl config --apiKey=xxx' or set RUNPOD_API_KEY")
```

The PR body says `ErrNoCredentials` is "shared by the rest client and the legacy api package so both paths report identically" — but there are **three** no-key branches, not two (`internal/api/client.go:43`, `api/query.go:43`, and this one). So:

```
runpodctl serverless list   -> {"error":"api key not found. get your key at …","code":"no_credentials"}
runpodctl ssh list-keys     -> {"error":"api key not found. run 'runpodctl config --apiKey=xxx' …","code":"cli_error"}
```

An agent branching on `code == "no_credentials"` to prompt for a key silently misses `ssh *`, `pod create`, `pod get`, `template create`, and `template update`. Two different messages for one condition, and this one still points at the deprecated `runpodctl config --apiKey` rather than `runpodctl doctor`.

### 2. `send` — only 3 of 6 failure paths were converted

The PR argues `send` is in scope ("a modern, documented command, not a legacy one") and fixes `GetFilesInfo`, `New`, and `Send`. But the four earlier exits in the same function are untouched ([`[cmd/transfer/transfer.go:79,84,86,92](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/cmd/transfer/transfer.go#L79)`](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/cmd/transfer/transfer.go#L79)):

```go
logger.Fatalf("file or folder %q does not exist", src)
```

So `runpodctl send /does/not/exist` still emits `runpod-send: file or folder "…" does not exist` — plaintext, no JSON, no `code`. Exit code is right, the shape isn't. `runReceive` is unchanged and still on `Run:` with the same five `Fatalf` sites. If send is in scope, finish the function; if it isn't, drop it and keep the PR narrower.

### 3. README states the contract as absolute, and it isn't yet

> `code` | always present.

That's not true for the `cobra.CheckErr` paths — legacy `cmd/pod`, `cmd/pods`, `cmd/cloud`, `cmd/ssh/commands.go`, and the ~18 remaining sites in `cmd/model/addModelToRepo.go` — nor for the `log.Fatal` paths in `cmd/croc` and `cmd/transfer`. Those emit `Error: <msg>` or `<prefix>: <msg>` in plaintext.

The PR acknowledges the `model add` subset, but the README is the artifact an agent author will trust, and it doesn't carry the caveat. Either scope the claim ("every error from the v2 resource commands…") or add a one-line known-gaps note. As written it will produce a JSON parse failure on a path someone reasonably expected to be covered.

### 4. `stockRank` can now report a literal "None" as available

[`[internal/api/gpu.go:205](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/internal/api/gpu.go#L205)`](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/internal/api/gpu.go#L205) — unknown non-empty now ranks 1, above absent's 0. Combined with:

```go
available := hasStock && stockStatus != ""
```

if the API's `stockStatus` enum ever returns a literal `"None"` / `"Unavailable"` (rather than an empty string), it ranks above absent, wins as the top-level `stockStatus`, and sets `available: true` for a GPU with no stock. Same class of wrong answer as the bug this change fixes, just in the other direction. `TestStockRank` only pins high/medium/low, so it wouldn't catch it.

**Question:** does `dataCenters.gpuAvailability.stockStatus` ever return a non-empty "no stock" literal? If yes, it needs an explicit entry in `stockOrder` ranked *below* `low`, not the unknown bucket.

Related, and this compounds it: the per-DC breakdown now substitutes `"none"` for an empty status, so `"stockStatus":"none"` means either "the API said none" or "the API said nothing and we filled it in" — the agent can't distinguish. You flagged the top-level/per-DC sentinel mismatch as a contract decision, which is fair, but this collision is new in this PR rather than pre-existing.

### 5. `modelRepoError` is now a no-op

[`[cmd/model/errors.go](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/cmd/model/errors.go)`](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/cmd/model/errors.go) — after the refactor, all three call sites reduce to `return err`:

```go
if mrErr := modelRepoError(err); mrErr != nil {
    return mrErr
}
return err
```

Branch 2 returns `err` unchanged (identical to the fallthrough), branch 3 returns nil (unreachable, since callers only enter on `err != nil`). The *only* behavioral difference left is branch 1, which returns the bare `ErrModelRepoNotImplemented` sentinel instead of the wrapped error — i.e. it silently strips context. Either delete the function and the wrapper blocks, or keep it and comment why discarding the wrapping is intentional.

### 6. `parseAPIError` can still put a raw JSON blob in `error`

[`[internal/api/client.go:238](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/internal/api/client.go#L238)`](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/internal/api/client.go#L238) — a non-2xx body that is valid JSON but carries neither `error` nor `message` falls through to `strings.TrimSpace(string(body))`:

```
{"detail":"validation failed","field":"gpuIds"}
  -> {"error":"{\"detail\":\"validation failed\",\"field\":\"gpuIds\"}","code":"bad_request","status":400}
```

Which is the double-encoding the PR exists to eliminate, and directly contradicts the README's "never a nested json blob." Worth extending the envelope with `detail`, or reusing `extractGraphQLMessage`'s approach, or at minimum falling back to `codeForStatus`'s english rather than the body.

### 7. `api/query.go` now returns a type reachable only from `internal/`

`github.com/runpod/runpodctl/api` is a non-`internal` package, so it's importable outside the module. Returning `internalapi.ErrNoCredentials` from `Query` means an external caller can no longer `errors.Is`/`errors.As` against it — they're back to string matching. The message text also changed from `"API key not found"`, which is a behavior change not listed in your "⚠️ Behavior changes" section. If `api` is meant to stay importable, export a sentinel there and have both paths wrap one shared value.

### 8. CI doesn't run the e2e suite — and this PR is the argument for fixing that

[`[.github/workflows/ci.yml:26](https://github.com/runpod/runpodctl/blob/main/.github/workflows/ci.yml#L26)`](https://github.com/runpod/runpodctl/blob/main/.github/workflows/ci.yml#L26) is `go test -vet=off --cover ./...` — no `-tags e2e`. The `pod create` global-networking regression in this PR was caught *only* because you ran the e2e suite by hand. That's concrete evidence CI would have shipped it. Worth a follow-up to wire `-tags e2e` in as a nightly at minimum, with the API key in secrets.

### 9. Smaller things

- **`urls` on `serverless update`:** [`[internal/api/endpoints.go:188](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/internal/api/endpoints.go#L188)`](https://github.com/runpod/runpodctl/blob/justinlin/con-683-685-runpodctl-agent-output/internal/api/endpoints.go#L188) populates `URLs` in `UpdateEndpoint` too, but the PR body and the README env-var table both say create/get/list. Fine behavior, stale docs.
- **`securePricePerHr,omitempty` on a `float64`:** 0 and "not reported" are indistinguishable, and `secureCloud: true` with the price key absent is reachable. For a field whose whole purpose is letting an agent cost-optimize, `*float64` with an explicit `null` is a cleaner contract than a key that vanishes.
- **Price field naming is split:** `GpuType` marshals `securePrice`/`communityPrice`; only `cmd/gpu`'s output struct renames to `securePricePerHr`. Anything that marshals `GpuTypeWithAvailability` directly emits the un-suffixed names.
- **`ErrNoCredentials` is a mutable package-level `*APIError`.** Any caller can scribble on `.Message`. A value + accessor, or at least a comment, would be safer.

### 10. Scope

65 files, +1322/−282, three tickets, plus an audit that reached `send`, `update`, `model`, and a full docs regeneration. The exit-code changes to `send`/`update`/`model` are script-visible behavior changes on commands with no relationship to CON-683/684/685. They're correct fixes and I'd merge them — but as their own PR they'd be reviewable in ten minutes and bisectable if someone's CI script breaks. Right now a `git revert` of the invoke-URL work also reverts the `send` exit code.

None of 1–6 look like blockers to me except at your discretion; 1 and 3 are the two I'd want resolved before this lands, since they're the ones where an agent trusting the documented contract gets a wrong answer rather than a rough one.

justinwlin:
Thanks — genuinely useful review. Everything addressed except where noted. #312 is also merged into this branch, so it's one PR now.

**Your two blockers — fixed**

1. `NewGraphQLClient` now returns the shared `ErrNoCredentials` sentinel — all five no-key paths (`pod list/create/get`, `ssh list-keys`, `template create`) emit `code: no_credentials`.
2. README no longer claims `code` is always present — added a known-gaps table (legacy pod commands, `exec`, and `project` — filed as CON-816).

**Also fixed**

- **2 — `send` half-done**: finished all six paths, and converted `receive` too (it had five more `Fatalf` sites). Both are `RunE` now: one JSON error object, exit 1.
- **4 — `stockRank` false-available**: literal `none`/`unavailable`/`out of stock`/`no stock` now rank 0, and `available` derives from `stockRank` via a shared `hasStock()`, so the two can't disagree. (To your question: I've only ever seen `High`/`Medium`/`Low` + empty from the API, so this is defensive.)
- **6 — `parseAPIError` nested blob**: reads `detail` too, and falls back to the status line instead of the raw body.
- **7 — `api/query.go` internal type**: `api.ErrNoCredentials` re-exported as the *same value*, so one `errors.Is` works for external callers. Message change added to behavior-changes list.
- **5 — `modelRepoError`**: deleted. Your read was exactly right.
- **9b — price `omitempty`**: now `*float64` with explicit `null`. Live case, not theoretical — six GPUs in today's `gpu list` hit it.
- **9a — `urls` on `serverless update`**: docs corrected.
- **9d — `ErrNoCredentials` mutability**: documented read-only, match with `errors.Is` (a value would break `errors.Is` identity).

**Deferred**

- **9c — price field naming split**: renaming the `internal/api` tags changes the wire shape; will do as its own change.
- **10 — splitting the exit-code fixes**: keeping them here (each was found auditing this PR's own contract); happy to split if you'd rather.
- **8 — e2e in CI**: agreed, most valuable finding. The `--global-networking` regression was only caught by manual `-tags e2e`. Say the word and I'll file a nightly-CI ticket.

justinwlin:
Follow-up on 9c, 8, and 10.

**9c (price naming) — tried the rename, backed it out: it's a silent data bug.** Those tags aren't output shape, they're the *decode* contract for the GraphQL response (which sends `securePrice`/`communityPrice`). Renaming them makes `json.Unmarshal` miss both fields, so every price silently becomes `0`. Caught before pushing, verified against live data. Left a `DO NOT rename` comment instead — nothing outside `internal/api` marshals GPU data, so users only ever see the `PerHr` names. A separate decode DTO would fix it cleanly if you'd rather have that.

**8 (e2e in CI) — staying manual for now, but the suite is green end to end for the first time.** ~7 tests provision real pods/endpoints, so per-PR runs cost money and a mid-test failure leaks a running pod. Two fixes along the way:

- Replaced a test pinned to a now-delisted Hub template with two contract-based tests.
- That surfaced a real bug: `template search` / `hub search` printed prose on stdout when nothing matched — the same stdout-contract violation as CON-683, on the success path. Both now emit `[]`.

If someone wires this into CI later: split by cost — read-only tests per-PR, resource-creating tests nightly. The `--global-networking` regression would have been caught by the free half.

**10 (splitting exit-code fixes) — keeping them here.** The repo squash-merges, so splitting doesn't buy revert granularity, and re-reviewing 20 commits costs more than a targeted forward-fix later. If it's a blocker for your approval rather than a preference, say so and I'll split.

LarveyOfficial (approved):

PR #313, opened 2026-07-28
fix: change int params in the EndpointupdateRequest struct

### LINKED ISSUE #298: runpodctl serverless update --workers-min 0 does not send workersMin in PATCH request
Expected behavior
workersMin should be updated to 0.

Actual behavior
The endpoint does not transition to workersMin=0 as expected.
In practice, 1 -> 2 and 2 -> 1 updates work, but 1 -> 0 does not.
Suspected root cause
From code inspection, the CLI command handling looks correct, because the update path accepts 0:
updateWorkersMin is checked with >= 0

However, the API request struct appears to use omitempty on int fields:
https://github.com/runpod/runpodctl/blob/171d9264a7f09738743015e6648e4ae841340562/internal/api/endpoints.go#L68

Because workersMin is an int with omitempty, setting it to 0 likely causes the field to be omitted entirely from the PATCH JSON body.
That would explain the exact behavior:
workersMin=2 is sent
workersMin=1 is sent
workersMin=0 is omitted
Suggested fix
Use pointer types for numeric PATCH fields so that 0 can be distinguished from "not provided", for example:

```
type EndpointUpdateRequest struct {
    Name        string `json:"name,omitempty"`
    WorkersMin  *int   `json:"workersMin,omitempty"`
    WorkersMax  *int   `json:"workersMax,omitempty"`
    IdleTimeout *int   `json:"idleTimeout,omitempty"`
    ScalerType  string `json:"scalerType,omitempty"`
    ScalerValue *int   `json:"scalerValue,omitempty"`
    Flashboot   *bool  `json:"flashboot,omitempty"`
}
```

dhollinger:
# Summary

Updated the following parameters in the api.EndpointUpdateRequest struct from type `int` to an integer pointer (`*int`).

* WorkersMin
* WorkersMax
* IdleTimeout
* ScalerValue

This is to fix a bug in the enpoints where a value of `0` causes the generated JSON to omit these fields as "empty" as per the `omitempty` JSON tag. Changing them to integer pointer (`*int`) will allow a `nil` value to trigger the `omitempty` JSON tag, while a value of `0` will be treated as a valid value.

fixes #298

Test plan:
- [x] go test ./...
- [ ] go test -tags e2e ./e2e

### DIFF
--- a/cmd/serverless/create.go
+++ b/cmd/serverless/create.go
@@ -246,8 +246,8 @@ func runCreate(cmd *cobra.Command, args []string) error {
 	}
 
 	input := &api.EndpointCreateGQLInput{
-		WorkersMin: createWorkersMin,
-		WorkersMax: createWorkersMax,
+		WorkersMin: &createWorkersMin,
+		WorkersMax: &createWorkersMax,
 	}
 
 	// compute type: gpu uses a gpu pool id, cpu uses an instance id.
@@ -308,7 +308,8 @@ func runCreate(cmd *cobra.Command, args []string) error {
 
 	if createExecutionTimeout >= 0 {
 		// 0 is allowed (server treats it as "no per-request limit").
-		input.ExecutionTimeoutMs = createExecutionTimeout * 1000
+		ms := createExecutionTimeout * 1000
+		input.ExecutionTimeoutMs = &ms
 	}
 
 	// flash boot maps to the flashBootType enum (off|flashboot); always set so
--- a/cmd/serverless/serverless_test.go
+++ b/cmd/serverless/serverless_test.go
@@ -3,7 +3,9 @@ package serverless
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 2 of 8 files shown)


justinwlin (approved):
Approved. Root cause was right — `*int` is the only way Go can tell "set to 0" from "not set" behind `omitempty`.

@dhollinger — I pushed a commit here via maintainer edits and rebased onto `main`. Your two commits are untouched. It adds:

- **Same bug on create.** `EndpointCreateGQLInput.WorkersMin/.WorkersMax/.ExecutionTimeoutMs` were still `int`+`omitempty`, so `create --workers-max 0` got the server default of 3. Now `*int`. `gpuCount` stays an int, API minimum is 1.
- **Output hid the fix.** The `Endpoint` response struct had `omitempty` too, and the CLI re-marshals it to print — a successful `--workers-min 0` showed no `workersMin` at all.
- **Validation** for `--idle-timeout` (1-3600) and `--scale-threshold` (>=1), matching create. Behavior change: both used to exit 0 and do nothing at 0, now they error.
- **Tests assert the request body.** The old test only checked the mocked response and passed either way. The other three fields could each be made a no-op with the suite green.

PR #316, opened 2026-07-29
feat(serverless): add health, run and status commands (CON-688)

justinwlin:
Adds `serverless health`, `serverless run` and `serverless status` so the deploy→verify loop stays in the CLI instead of dropping to curl with a hand-pasted API key.

Linear: [CON-688](https://linear.app/runpod/issue/CON-688)

## What changed

- **`serverless health <id>`** — wraps `GET /v2/<id>/health`, body passed through verbatim.
- **`serverless run <id> --input '{...}'`** — submits the job, polls until terminal, prints the job payload. `--input-file <path>` and `--input -` (stdin) for payloads that fight shell quoting; json is validated locally so a quoting mistake is a local `usage_error`, not an API round trip.
- **`serverless status <id> <job-id>`** — the follow-up for `--no-wait` and for a run that hit its `--wait` budget.
- **`internal/api.InvokeClient`** — `api.runpod.ai/v2` is a different service from the control plane, so it gets its own client with per-call deadlines; auth, user-agent and error parsing shared so codes stay identical.
- **Two new error codes** (`timeout`, `job_failed`), registered in `client.go`, the README table and AGENTS.md; `timeout` also applied to `model add --wait-for-hash` so the code means one thing CLI-wide.

## Notable decisions

- **`run` uses `/run` + `/status` polling, never `/runsync`** (the ticket said `/runsync` by default). `/runsync` is not synchronous: it holds the connection ~90s then answers with a still-running job, returns no job id until it answers (a timed-out request strands a billed, unpollable job), and `sync-` job results expire in 1 minute vs 30 for `/run`. All three legs source-verified in `ai-api`. Consequently there is no `--async` flag — that behavior is the only behavior.
- **`--wait` defaults to 5m** (measured ~95s cold CPU start; ai-api's own `?wait` ceiling is 5m). `--no-wait` ≡ `--wait 0`. Per-call timeout is clamped to the remaining budget, with a 1s floor so the last poll can still return an answer.
- **Exit/stream discipline:** a `FAILED` job exits 1, but the worker's traceback is data — payload on stdout, coded error on stderr. Same on `timeout`, where the error names the `serverless status` follow-up.
- **`policy`/`webhook`/`s3Config` are not exposed** (scope cut, follow-up below), but a pasted curl envelope now triggers a warning naming the ignored keys instead of being silently double-wrapped.

<details><summary>Full deviation analysis and review triage (2 rounds, 26 findings: 25 fixed, 1 rejected)</summary>

Ten deviations from the ticket as written, each source-verified by an independent reviewer against `ai-api` (rulings: 7 valid, 2 valid-but-understated, 1 fixed-in-round). Highlights beyond the decisions above:

- `serverless status` is a third command the ticket didn't name — but both non-terminal outcomes print a message referencing it; shipping the messages without the command is a dangling reference back to curl. Ruled load-bearing.
- Non-object `--input` (arrays, scalars, `null`) rejected locally before any API call; matches ai-api's own 400.
- `timeout` is emitted for plain per-call deadlines too (e.g. `health` against a stalled host), where nothing is left running; the README now distinguishes the two cases rather than the code being narrowed (narrowing would thread wait-awareness through `do()` for little gain).
- The ticket's "ship `health` only" smaller cut was declined: `health` answers "is the endpoint up", not "does my handler work", and `run` is what the auth-friction problem is about.

Key review fixes: dropping `/runsync` at the design level (billed unpollable jobs); payload fidelity (64-bit ints survived; handler's `gpuTypeId` no longer renamed); job id captured once instead of re-read from poll responses (a status-only body used to blank it → `GET /status/` → spurious `not_found`); poll clamping to the wait budget now tested (a mutation had `--wait 3s` running 26.5s with tests green); retry backoff unified (429 hammering at 2 req/s → shared `nextPollInterval`); `--no-wait` on a 200 with no job id exits 1 instead of silently succeeding.

Rejected (1): "`--wait` may overshoot by up to 1s" — the 1s floor is deliberate; clamping the last poll to a few ms guarantees a timeout instead of an answer. The README/help now state the actual bound.

</details>

## Testing

Unit: table-driven, `httptest` for the invoke client, injected mock for the commands — payload fidelity, envelope detection, exit-code mapping, retry policy, deadline clamping in both directions, stdout/stderr discipline. `cmd/serverless` 74%. Gates: `gofmt`/`go vet` clean, `go test ./...` all 15 packages, docs-gen no diff.

<details><summary>Evidence: stub-host repros, live prod e2e with cleanup proof</summary>

Stub invoke host (`RUNPOD_INVOKE_URL`, free):

```
# job id survives a poll response that drops it (was: 3 polls to /ep1/status/ )
$ serverless run ep1 --input '{}' --wait 3s
GET /ep1/status/job-d x4 -> {"error":"gave up waiting for job job-d ... poll it with: runpodctl serverless status ep1 job-d","code":"timeout"}

# 429 backoff: 11 requests in 5s -> 6
# --no-wait, submit answers with no job id: exit 1, api_error (was exit 0, silent)
# curl envelope: note names the ignored policy/webhook/s3Config keys
# --input null / unreadable --input-file: local usage_error
```

Live prod e2e — throwaway CPU endpoint on `runpod/mock-worker:latest`, deleted via `t.Cleanup`:

```
--- PASS: TestE2E_ServerlessInvokeLifecycle (88.95s)
    /run_waits_for_the_result (51.67s)            # cold start, /run + /status polling
    /no-wait_returns_the_job_id_then_status_follows_it
    /input_from_stdin
    /failed_job_exits_1_with_the_payload_on_stdout
    /wait_budget_too_small_is_an_actionable_timeout  # runs the suggested command verbatim
--- PASS: TestE2E_ServerlessHealth / ...NotFound / ...RejectsBadInputLocally
```

Payload fidelity before/after: `"seed": 12345678901234567000` → `12345678901234567890`; `"gpuId"` → `"gpuTypeId"` preserved. Cleanup verified: `serverless list` back to the 6 pre-existing endpoints, both e2e templates 404. (Five pre-existing `cli_test.go` failures are unrelated — stale shared binary + prod data drift.)

</details>

## Follow-ups

- [ ] update the runpodctl agent skill — `serverless health/run/status`, the `timeout`/`job_failed` codes (incl. `model add --wait-for-hash` moving from `cli_error` to `timeout`), the "stdout carries the payload even on failure" rule
- [ ] expose `--policy-ttl` / `--execution-timeout`, `webhook`, `s3Config` (most needed when raising `executionTimeout` to match a long `--wait`)
- [ ] `serverless cancel <job-id>` and `/stream` support
- [ ] `health` answered normally for a just-deleted endpoint (invoke-side caching) — don't treat it as an existence check


lukepiette (approved):
Approving. The new commands are genuinely additive — no command/flag/alias collisions, no existing help text changed beyond the generated docs index — and the risky decisions check out: polling is deadline-bounded with clamped per-call timeouts (cannot spin forever, pinned by `TestBoundedRequestTimeout` and the loop-level budget test), `/run` over `/runsync` is right given sync-result expiry, credentials and host conventions reuse the existing `RUNPOD_INVOKE_URL`/config plumbing, stdout/stderr discipline is clean (payload on stdout even for FAILED, progress on stderr), and `output.PrintRaw` is genuinely needed to avoid the normalizer corrupting handler payloads. Test suite is high-signal; nothing pre-existing weakened.

Comments (none blocking):

1. **One real contract change is hiding under a feature title:** `model add --wait-for-hash` now emits error code `timeout` where it emitted `cli_error`. Exit code and message text unchanged, intentional and tested — but scripts branching on the stderr JSON `code` field break, so it needs a release-note line.
2. **The `timeout` code is machine-ambiguous.** It covers both "wait budget exhausted, job still running, do NOT re-invoke" and "single call timed out, retry is safe," distinguishable only by message prose — which forces exactly the message-parsing the repo's error-object design exists to avoid. A `retryable: true/false` field or a second code would be cleaner; fine as a follow-up. Related inconsistency: `fetchJobStatus` returns the raw last error when the budget is consumed by transient failures, where `waitForTerminal` special-cases it.
3. Minor: `--no-wait --wait 10m` is accepted with the wait silently ignored (one `Usagef` fixes it); `resolveJobInput` never checks payload size, so an oversized body costs a full upload before the API's 4xx (a local check would match the PR's own fail-locally philosophy); endpoint/job ids are not `url.PathEscape`d in `/status`/`/health` paths.
4. **Coordinate land order with #317:** both PRs touch the error-code doc block in `internal/api/client.go`, `internal/output/output.go`, and both add a `/health` client (this PR's raw `InvokeClient.EndpointHealth` vs #317's typed `Client.GetEndpointHealth`). No compile collision, but whichever lands second should actually consolidate the duplicate, as #317's body already promises.

PR #317, opened 2026-07-29
feat(pod,serverless): add --wait and --wait-timeout to create (CON-689)

justinwlin:
Adds an opt-in `--wait` to `pod create` and `serverless create` so create returns when the resource is *usable*, not when it is scheduled — today a pod reports `RUNNING` minutes before sshd answers, so agents sit in a poll loop guessing.

Linear: [CON-689](https://linear.app/runpod/issue/CON-689)

## What changed

- `pod create --wait` returns when the pod's public port 22 accepts a TCP connection **and** answers with an SSH banner, then prints the `pod get` shape (so the payload carries the live `ssh` block).
- `serverless create --wait` returns when `/health` reports a `ready` or `running` worker.
- `--wait-timeout` (default `10m`; reuses the existing `7d`-aware duration parser, now shared in `internal/duration` and range-checked).
- New `internal/waitfor`: one bounded poll loop with an injectable clock, throttled stderr progress, and a typed error carrying the last known state. Progress is a caller-supplied `io.Writer`, so the loop structurally cannot touch stdout.
- The legacy `PodSSHConnection` loop (`cmd/project/ssh.go`) now calls the shared helper — its stdout line, wording and timings are byte-identical to main (diff in details), and it loses a bug where a poll that succeeded at the deadline still reported timeout.
- Error objects gained an optional `id` field, so a `wait_timeout` names the billed resource as **data**, not prose. Two new codes: `wait_timeout`, `wait_interrupted`.

## Notable decisions

- **SSH-ready means a real banner, not port presence.** The ticket's implementation note assumed `runtime.ports` means reachable; it doesn't — docker publishes the binding at container start regardless of whether anything listens (measured: public port 22 in ~25s while `nc` was refused). The legacy loop was still extracted and reused as asked. No handshake/key check, so `--wait` can succeed on an image whose sshd never got your key — disclosed in help and README.
- **Endpoint-ready is `/health` `ready > 0 || running > 0`.** Neither counter is as strong as "a worker is ready" (`ready` counts a flashboot-cached EXITED worker; `running` is written at scheduling time), but `/health` exposes nothing stronger, and `GetEndpoint(includeWorkers)` returns historical records. The success line names which counter fired. Backend ask filed as follow-up.
- **`--workers-min 0` warns instead of refusing.** An earlier refusal rested on a false premise (review-refuted at the source: standby workers are floored to 5 whenever `workersMax > 1`, and the CLI's own polling drives provisioning) — live probe showed 6 endpoints with `workersMin` unset all reporting `ready > 0`.
- **Transient poll failures don't end a wait** — the resource already exists and bills, and a transport code would tell an agent to retry the create and buy a second one. Fatal only: HTTP 400/401/403, a terminal pod status, or the resource absent from two consecutive reads.
- **Nothing is deleted on timeout or ctrl-c.** The error carries the id, last known state and the delete command; deleting a resource the user didn't ask to delete is worse than leaking one you were told about.

<details><summary>Full deviation analysis and review triage (2 rounds, 28 findings: 20 fixed, 1 rejected, rest disclosed/nits)</summary>

Ten deviations from CON-689's wording, each re-derived by an independent reviewer from `ai-api`/`runpod-backend`/`host` source or free live probes. Beyond the decisions above:

- `pod create --wait` prints the `pod get` shape, not the create response — neither create response carries an ssh command, and a connectable pod is the flag's purpose. The re-read now passes `includeMachine=true`; the residual `env`/`ports` type difference vs graphql is disclosed.
- `--compute-type CPU` and community-without-`--public-ip` warn instead of failing (satisfiable but often not); only `--ssh=false` is refused (genuinely unsatisfiable).

Key review fixes: auth fail-fast was unreachable on the pod path (all graphql failures carried `graphql_error`; a bad key burned the full budget — now 0.4s fatal via HTTP status); a deleted endpoint burned the whole 10m (a `/health` 404 after a successful read is now fatal in ~10s; 404 before first read and any 5xx stay transient); `FindPodConnection` nil-deref on a `pods:[null]` element right after `--wait` success (SIGSEGV → nil guard); `duration.Parse` silently going negative at `>= 106752d` (now range-checked, also fixes `pod list --since 200000d`); one transient empty `myself{pods}` read ending the wait as "terminated" (now two consecutive misses); e2e cleanup registered before each create so `t.Fatalf` paths can't orphan a billed resource.

Rejected (1): a "blocker" mutation report where a concurrent process had left an edit in the shared worktree mid-review; moot — that guard is gone.

Nits later fixed in `16e96ba`: unused `State.Err` deleted, two unreachable fallback branches removed, and the defaults test strengthened so `DefaultTimeout`/`DefaultInterval`/`DefaultProgressEvery` no longer survive mutation. Still skipped: legacy 5s account-wide poll, re-read backoff ignoring the wait context (~4s worst-case overshoot, inherited from the legacy loop).

</details>

## Testing

`internal/waitfor` 95%+, `internal/duration` 100%. Gates: `gofmt`/`go vet` (incl. e2e tags) clean, `go test ./...` all 16 packages, docs regenerated. Every substantive review finding was reproduced through the real binary against a fake control plane before fixing, and re-run after.

<details><summary>Evidence: fake-control-plane repros, live prod e2e with cleanup proof, legacy byte-diff</summary>

Fake control plane (free):

```
# graphql 401 mid-wait:            15.05s wait_timeout  ->  0.41s graphql_error, status 401, id present
# pods:[null] after wait success:  panic SIGSEGV, exit 2  ->  exit 0, one parseable json doc
# one transient empty pods read:   ended wait as not_found  ->  succeeds on poll 3; genuine deletion still fails fast (2 misses)
# endpoint deleted mid-wait:       burned full 90s  ->  10.48s not_found with id
# --wait-timeout 200000d:          silently waited 10m default  ->  rejected: out of range (max 106751d)
```

Live prod e2e (round 1; paid paths not re-bought — round 2 verification was free: source reads, read-only probes, fake plane):

```
timeout path:  exit 1, stdout 0 bytes, {"code":"wait_timeout","id":"kasgak1hp6sluk"} -> deleted via the id field
ctrl-c path:   exit 1, {"code":"wait_interrupted","id":"ekchrd6mlssk98"} -> deleted
success (pod): exit 0, ready after 12s, stdout is exactly one parseable object with a live ssh_command
success (ep):  exit 0, ready worker after 25s
leak check:    pod list -a -> []; serverless list -> same 6 pre-existing; templates unchanged
```

Legacy `project`/`exec` behavior, base vs head against an unreachable graphql: stdout and stderr byte-identical.

</details>

## Follow-ups

- [ ] update the runpodctl agent skill — `--wait`/`--wait-timeout` semantics, what `ready` actually proves, the `pod get` output shape, the new codes + `id` field
- [ ] ask the backend for a readiness signal that means "a container is up and can take a job" — neither `/health` counter does; today's predicate is the strongest thing a client can observe
- [ ] `internal/api.GetEndpointHealth` may collide with CON-688's `serverless health`; whichever lands second should reuse the other


lukepiette (approved):
Approving. The part that matters most on a flag-addition PR — the default path — is verified byte-compatible: all 147 deletions decompose into helper extractions and gofmt-only hunks (each diffed against main), no-wait stdout/exit codes/non-blocking behavior are unchanged for both creates, and the shared-helper refactors (`sshconnect`, `project ssh`, `internal/api/client.go` split) preserve caller behavior — including quietly fixing a real main bug where a poll succeeding at the deadline still reported timeout, and a nil-pod SIGSEGV. The pod-ready predicate (RUNNING + port 22 mapped + TCP connect + `SSH-` banner) is the right strength, keeping the billed resource on timeout with a machine-readable `id` is the right call, and the fake-clock tests mean none of this sleeps in CI.

Comments (none blocking):

1. **Ctrl-C window after readiness:** `defer stop()` in `waitForPodSSH` deregisters the signal handler before `podDetailsWithSSH` runs its up-to-~4s of context-blind re-read sleeps. A Ctrl-C in that window exits 130 with empty stdout — no JSON error object, no `id` — the one path where the pod is orphaned as far as machine-readable output goes. Keeping the signal ctx alive through the re-read (and threading it into the backoff sleeps) closes it.
2. **Asymmetric fail-fast:** an endpoint deleted before `/health` ever answers 404s "transiently" forever and burns the full 10m budget, because the resource-gone fail-fast only arms after one successful read. Narrow corner, but it is asymmetric with the pod path's two-consecutive-miss rule.
3. The endpoint-ready predicate (`ready > 0 || running > 0`) is soft since `running` is written at scheduling time — disclosed in help/README with a backend follow-up filed; holding that follow-up to account is worth doing.
4. **Coordinate land order with #315/#316:** #315 also refactors `cmd/pod/get.go` and `internal/sshconnect/sshconnect.go` (real merge conflicts, not just textual), and #316 adds the sibling raw `/health` client this PR's body promises to consolidate with. Landing #315 → #317 → #316 (or agreeing on any explicit order) beats three independent rebases onto moved code.

PR #318, opened 2026-07-30
docs: correct output-format, exec exit code, and skill path

justinwlin:
## what

Three README/AGENTS.md claims that don't match the code. Docs only — no code change.

| claim | reality |
|---|---|
| `--output=table   # human-readable table` | **no table format exists.** `internal/output/output.go` `ParseFormat` maps `yaml` and sends everything else to json, so `--output=table` silently returns json |
| `exec` documented as "plaintext, exit 1" | `cmd/exec/commands.go` uses `Run`, not `RunE` — the error goes to stderr and the process still **exits 0**. It also prints progress to **stdout** and polls `maxPollTime` (5 min, `cmd/project/ssh.go`) before failing |
| skill lives at `github.com/runpod/skills/tree/main/runpodctl` | repo renamed to `runpod-plugins-official`; the skill moved to `plugins/runpod/skills/runpodctl` |

<details>
<summary>verified against a build of this branch's base</summary>

```
$ runpodctl gpu list --output=table
[ { "available": true, ...          # json, exit 0, no warning

$ runpodctl exec python /etc/hosts --pod_id bogus-pod-id
warning: 'runpodctl exec' is deprecated; ...
Running remote Python shell...      # stdout
Waiting for Pod to come online...   # stdout, then polls 5 min
```

`grep -rn '"table"' --include='*.go' .` returns nothing, and `cmd/root.go` registers the flag as `output format (json, yaml)`.

</details>

## why now

Found while updating the runpodctl agent skill for the v2.8.0 output changes (runpod/runpod-plugins-official#41, CON-752). That work was supposed to use this README as its source, so the skill briefly inherited the `--output=table` error. The skill now documents the verified behavior; this PR removes the divergence rather than leaving the two to disagree.

CON-752

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,4 +24,4 @@ This document exists for non-obvious, error-prone shortcomings in the codebase, 
 - default output format is json (for agent consumption); never change this default.
 - e2e tests cost real money — always clean up resources with `t.Cleanup`.
 - always e2e test cli changes before considering work done: build the binary, run the new/changed commands against the live api (`RUNPOD_API_KEY` is in the env), and clean up any created resources immediately after verifying the response. this is non-negotiable.
-- keep the runpodctl skill in sync: https://github.com/runpod/skills/tree/main/runpodctl — update it whenever commands, flags, or behavior change.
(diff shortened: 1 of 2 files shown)


justinwlin:
Folded into #319, which now carries both the doc corrections and the `--output` validation fix against `main`. Closing to keep it to one review.

PR #319, opened 2026-07-30
fix: --output validation and exec exit code, plus README/AGENTS.md corrections

justinwlin:
## what

Two related things, both found while updating the runpodctl agent skill for the v2.8.0 output changes (runpod/runpod-plugins-official#41, CON-752): three doc claims that didn't match the code, and the flag behavior that let one of them go unnoticed.

### 1. `--output` was case-sensitive and swallowed unknown values

`--output` matched its value case-sensitively and sent anything unrecognized to json, so **`--output=YAML` and `--output=table` both silently returned json**. That silence is how a table format that never existed ended up in the README.

| before | after |
|---|---|
| `--output=YAML` → json | yaml |
| `--output=table` → json, exit 0 | `usage_error`, exit 1 |
| `--output=jsonl` → json, exit 0 | `usage_error`, exit 1 |
| `--output=yaml` / omitted | unchanged |

```jsonc
// runpodctl gpu list --output=table
{"error":"invalid --output \"table\": supported formats are json and yaml","code":"usage_error"}
```
…followed by usage text, the same shape a bad flag *name* already produced.

**How:** `ParseFormat` normalizes case + surrounding whitespace; a new `ValidateFormat` runs once in `rootCmd.PersistentPreRunE`, wrapped in `usageError` so classification comes from the type rather than a message prefix (per the `usageErrorPrefixes` note in AGENTS.md).

`ParseFormat` still defaults instead of erroring — it has **47 call sites** with no way to report one, and validation has already run before any of them execute. That keeps the diff to two files instead of fifty.

**Cobra caveat:** by default only the closest `PersistentPreRun(E)` runs, so `exec` and `config` — which define their own for a deprecation notice — shadowed this guard and ran their bodies with an unvalidated value. On `config` that saved the api key and generated + uploaded an ssh key on an invocation that should have exited 1. Caught in review by @brosenpod. Fixed by `cobra.EnableTraverseRunHooks = true`, which runs the root hook first for every command in the tree, then the subcommand's own. `TestOutputValidationSurvivesShadowingHook` pins both halves.

Paths that emit no data stay exempt, since cobra resolves them before any hook: `--help`, `help <cmd>`, a bare parent, `--version`/`-v`, shell completion. `help <cmd>` is the one that needed an explicit carve-out — it runs as an ordinary command, so the guard was blocking the help that lists the valid values.

`initConfig` runs from `cobra.OnInitialize`, i.e. ahead of the hook chain, so "rejected" doesn't mean "wrote nothing". It creates a default `~/.runpod/config.toml`, and because `config` binds `--apiKey` into viper at package init, a rejected `config --apiKey=… -o table` **still persists that key**. Validation can't prevent that from a hook — only moving `initConfig` after the chain would. What it does prevent is the `config` body: no ssh keypair generated or uploaded, and a readable existing config left alone.

### 2. Three doc claims corrected

| claim | reality |
|---|---|
| `--output=table   # human-readable table` | no such format (fixed above; the README line is gone) |
| `exec` documented as "plaintext, exit 1" | it used `Run`, not `RunE`, so the error went to stderr and the process **exited 0** — now fixed rather than just documented, see §3 |
| skill lives at `github.com/runpod/skills/tree/main/runpodctl` | repo renamed to `runpod-plugins-official`; skill moved to `plugins/runpod/skills/runpodctl` |

### 3. `exec` now reports failure

`runpodctl exec python` printed its error and exited 0, so a caller chaining on success ran anyway:

```bash
runpodctl exec python train.py && ./deploy.sh   # deployed on a failed run
```

`Run` has no way to return an error. Switched to `RunE` and return it — nothing deeper needed changing, since `PodSSHConnection` and `PythonOverSSH` already return errors instead of exiting.

| | before | after |
|---|---|---|
| failed run | plaintext on stderr, **exit 0** | flat json with a `code`, **exit 1** |
| stdout progress, deprecation notice, 5-min `maxPollTime` | | unchanged |

Progress stays on stdout deliberately: AGENTS.md requires legacy commands preserve their stdout exactly, and that output is progress, not an error.

This has no ticket — the README used to attribute it to CON-816, but that only ever covered `project` and is now closed as a duplicate of CON-874.

## compatibility

Behavior change: `--output=<junk>` now exits 1 rather than returning json. Anything passing `json`/`yaml` is unaffected, and the default when the flag is omitted is unchanged (per the "never change the default output format" constraint). Worth a minor-version note.

<details>
<summary>e2e against the live API, per AGENTS.md</summary>

```
rejected (exit 1, code usage_error):  --output=table, yml, jsonl, YAMLX, -o nope
                                      also `exec python -o table` and
                                      `config -o jsonl --apiKey=fake` — both now
                                      fail before their body runs (no ssh key
                                      generated; see the initConfig note above
                                      for what still gets written)
accepted:                             --output=json, yaml, YAML, Yaml, -o YAML
                                      `exec python -o yaml` still prints its
                                      deprecation notice, so traversing adds the
                                      root hook without dropping the child's
default (no flag):                    unchanged json
unaffected:                           version, legacy `get pod` (still a table),
                                      `serverless list -o yaml`
```

```
exec python /tmp/nope.py (no ssh key):  exit 1, stdout keeps "Running remote Python
                                        shell...", stderr gets the deprecation
                                        notice then {"code":"cli_error"}
```

`go build ./...`, `go vet ./...` and `go test ./...` all clean; new unit tests in `internal/output/output_test.go`, `cmd/root_test.go` and `cmd/exec/commands_test.go`. Mutating paths (`config`, `exec`) were exercised under a throwaway `HOME`. No resources created, so nothing to clean up.

</details>

CON-752


cmd/root.go
```diff
@@ -86,6 +86,23 @@ func init() {
...
+	//
+	// NOTE: cobra runs only the closest PersistentPreRun(E), so a subcommand that
+	// defines its own shadows this. `exec` and `config` do; neither honors
+	// --output, so nothing is lost, but a new subcommand that defines one must
+	// call output.ValidateFormat itself. TestOutputFlagValidationCoverage pins the
+	// current set so a new shadowing command shows up as a test failure.
+	rootCmd.PersistentPreRunE = func(c *cobra.Command, _ []string) error {
+		if err := output.ValidateFormat(outputFormat); err != nil {
+			return &usageError{cmd: c, err: err}
+		}
```
brosenpod: Cobra skips this root hook when a descendant defines PersistentPreRun(E). Consequently, `exec python` and `config` accept inherited invalid values such as `--output=table`; `config` can proceed with configuration and SSH mutations instead of returning `usage_error`, contrary to the stated behavior.

justinwlin: Confirmed, and worse — `config`'s body saves the api key *and* generates and uploads an ssh keypair, so `config --apiKey=k --output=table` did all of that on an invocation that should have exited 1.

Fixed in f31eb05 with `cobra.EnableTraverseRunHooks = true`: root hook runs first for every command, `exec`/`config` keep their own notices.

One correction to my own framing: rejection does **not** stop the key being saved. `initConfig` runs from `cobra.OnInitialize`, ahead of the hook chain, and `config.go:114` binds `--apiKey` into viper at init — so the key still lands in a new `~/.runpod/config.toml` on exit 1. Only the keypair half is prevented. Now documented.

Also carved out `help <cmd>`, which the guard was wrongly rejecting, and the test now drives the real `exec`/`config` instead of a stand-in.

Scope heads-up: this PR also fixes `exec`'s exit code now (used `Run`, so failures exited 0).

lukepiette (approved):

PR #320, opened 2026-07-30
fix: STO-330: Correctly mirror HF models for Model Repo

brosenpod:
Utilize newly added GQL for mirroring models from Huggingface to Model Repo

### DIFF
--- a/cmd/model/addModelToRepo.go
+++ b/cmd/model/addModelToRepo.go
@@ -27,6 +27,7 @@ import (
 var (
 	addModelOwner               string
 	addModelName                string
+	addModelHuggingFaceModel    string
 	addModelCredentialReference string
 	addModelCredentialType      string
 	addModelStatus              string
@@ -158,7 +159,9 @@ var addCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(0),
 	Short: "add a model",
 	Long:  "add a model to the runpod model repository",
-	RunE:  runAddModel,
+	Example: `  # --name is the destination; --huggingface-model is the source
+  runpodctl model add --name tiny-llm --huggingface-model arnir0/Tiny-LLM`,
+	RunE: runAddModel,
 }
 
 var AddModelToRepoCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


cmd/model/addModelToRepo.go
```diff
@@ -363,6 +376,18 @@ func runAddModel(cmd *cobra.Command, args []string) error {
 	return printModelAddOutput(cmd, modelAddOutput{Model: model, Upload: result.Upload})
 }
 
+func validateAddModelFlags() error {
+	if !isHuggingFaceMirror() {
+		return nil
+	}
+
+	if addModelDirectoryPath != "" || addModelCreateUpload || addModelFileName != "" || addModelFileSize != "" || addModelPartSize != "" || addModelContentType != "" || addModelWaitForHash {
```
jebenexer, line 384: Nit: no client-side shape check on `--huggingface-model` (expects `owner/repo`) before this hits the network. The server already validates it correctly (`sourceParts.length < 2` in the `addModelToRepo` resolver), so this is just a UX nit — a malformed value round-trips to the API instead of failing locally. Not blocking.

docs/runpodctl_model_remove.md
jebenexer, line 1: Question: this file (and `docs/runpodctl_serverless_update.md`) only change by a trailing-newline removal, unrelated to this PR's feature. Assuming that's a side effect of regenerating all docs — just flagging so it's not mistaken for an unrelated manual edit. Not blocking.

cmd/model/addModelToRepo.go
```diff
@@ -195,18 +199,26 @@ func bindAddModelFlags(cmd *cobra.Command) {
...
+}
+
+// wantsUploadSession reports whether the flags ask for a local upload session.
+// Metadata retains its legacy upload behavior except when it belongs to a
+// server-side Hugging Face mirror.
 func wantsUploadSession() bool {
 	return addModelCreateUpload || addModelFileName != "" || addModelFileSize != "" ||
-		addModelPartSize != "" || addModelContentType != "" || len(addModelMetadata) > 0
+		addModelPartSize != "" || addModelContentType != "" ||
+		(len(addModelMetadata) > 0 && !isHuggingFaceMirror())
```
jebenexer, line 212: Follow-up (non-blocking): `wantsUploadSession()` here and the inline `isUploadFlow` computed a bit further down in `runAddModel` (hand-listing the same flags minus the metadata carve-out) are two independently-maintained predicates over the same flag set. I traced every reachable state and both stay correct today because `validateAddModelFlags()` rejects the mirror+upload-flag combos before either is used — but the comment above `wantsUploadSession` already warns this exact split "previously differed by `--metadata`" once before. Worth a small follow-up to derive `isUploadFlow` from `wantsUploadSession()` instead of a second hand-written list, so they can't drift again when a new flag is added.

jebenexer (approved):
Local review (checked out `pr-320` against current `origin/main`, ran full build/vet/test suite — all pass). Cross-verified the `huggingFaceModel` field and provider-forcing behavior against the actual backend resolver on `runpod/RunPod` main (`node/graphql/schema/modelRepo.ts`) — wire-compatible.

No blockers. Strong test coverage on the new mirror path, including every blocked flag-combination and regression tests for the two existing local-upload paths. Left 3 non-blocking comments (1 maintainability follow-up, 2 nits/questions). Approving — safe to merge.

PR #321, opened 2026-07-30
fix: sign and notarize macos binary (CON-787)

justinwlin:
macOS binaries ship with only the ad-hoc signature the Go linker emits — `TeamIdentifier` unset, `Signature=adhoc`, `spctl` verdict `rejected`. Gatekeeper refuses to run them; the brew cask path fails loudest, as *"the application is damaged and can't be opened."*

All four macOS install paths (brew formula, brew cask, `install.sh`, `runpodctl update`) fetch `runpodctl-darwin-all.tar.gz`, so signing that one artifact fixes all four.

## Changes

- `.goreleaser.yml` — `quill sign-and-notarize` post-hook on `universal_binaries`, after `lipo` merges the per-arch builds. `output: true` so quill's notary submission id and rejection reasons reach the job log; GoReleaser swallows hook stdout otherwise.
- `release.yml` — install quill (pinned to `v0.7.1`), pass five `QUILL_*` secrets to GoReleaser, and bound the job at `timeout-minutes: 45`.

Snapshots ad-hoc sign and skip notary submission (`--dry-run={{ .IsSnapshot }}`), so local builds need no credentials. Job stays on `ubuntu-latest` — quill signs and notarizes Mach-O from Linux.

The timeout matters because quill waits for a conclusive notary verdict with no bound of its own: a stalled Apple submission would otherwise hold the `release-${ref}` concurrency lock for the default 6h and block any retry.

## Verification

<details>
<summary>Snapshot build passes; hooks inherit job env; the real universal binary signs valid against the production cert</summary>

`goreleaser release --snapshot --clean` → exit 0, hook fires on the merged binary and its output is now visible:

```
• universal binaries
  • creating from 2 binaries    id=default binary=dist/default_darwin_all/runpodctl
  • running hook                hook=quill sign-and-notarize "dist/default_darwin_all/runpodctl" --dry-run=true --ad-hoc=true
    │ [0000]  WARN only ad-hoc signing ...
    │ [0000]  WARN DRY RUN: skipping notarization...
```

GoReleaser hooks inherit the job environment — verified with a sentinel variable in an isolated project, since that is how the five secrets reach quill:

```
• running hook   hook=sh -c 'echo "HOOK SEES SENTINEL=[$SENTINEL_VAR]"'
  │ HOOK SEES SENTINEL=[leaked-into-hook]
```

Same universal artifact signed with the real Developer ID `.p12`:

```
Authority=Developer ID Application: Runpod, Inc. (YM4PQ356FQ)
Authority=Developer ID Certification Authority
Authority=Apple Root CA
TeamIdentifier=YM4PQ356FQ

codesign --verify --deep --strict:
  valid on disk
  satisfies its Designated Requirement
```

Not yet exercised: the live Apple notary round-trip, which only runs on a real tag. Watch the first tagged release.

</details>

## Notes

- Signing is not retroactive — existing v2.3.0 artifacts stay unsigned. A new tag is required after merge.
- Certificate is Developer ID Application, G2 Sub-CA, valid to 2031-07-31. Previous Sub-CA certs would have expired 2027-02-01.
- Fails closed: notarization failure aborts during the build stage, before the GitHub release or tap PR exist. No half-published release.
- `upx` stays scoped to `linux_amd64` — it rewrites the Mach-O and would destroy the signature.
- Bare binaries in `.tar.gz` can't be stapled (`xcrun stapler` is `.app`/`.dmg`/`.pkg` only), so Gatekeeper does an online ticket lookup. Expected.
- The `legacy` per-arch darwin binaries remain unsigned. Only pre-change clients self-update via those; current `update.go` and `install.sh` both use `darwin-all`. Unchanged by this PR.

Does not fix: Homebrew tap trust (`HOMEBREW_REQUIRE_TAP_TRUST`) or self-update checksum verification ([API-318](https://linear.app/runpod/issue/API-318)). Both orthogonal.

[CON-787](https://linear.app/runpod/issue/CON-787/fix-runpodctl-homebrew-install-on-macos)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,6 +14,10 @@ jobs:
     name: release
     runs-on: ubuntu-latest
     concurrency: release-${{ github.ref }}
+    # quill waits for a conclusive notary verdict with no bound of its own;
+    # without this a stalled apple submission holds the concurrency lock for
+    # the default 6h and blocks any retry
+    timeout-minutes: 45
 
     steps:
       - name: Checkout
@@ -36,8 +40,10 @@ jobs:
         with:
           go-version: "1.26.5"
 
+      # pinned: an unpinned installer makes releases non-reproducible and lets
+      # an upstream regression break signing without a change on our side
       - name: Install quill
-        run: curl -sSfL https://get.anchore.io/quill | sudo sh -s -- -b /usr/local/bin
+        run: curl -sSfL https://get.anchore.io/quill | sudo sh -s -- -b /usr/local/bin v0.7.1
 
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v6
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,6 +36,9 @@ jobs:
         with:
           go-version: "1.26.5"
 
+      - name: Install quill
+        run: curl -sSfL https://get.anchore.io/quill | sudo sh -s -- -b /usr/local/bin
+
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v6
         with:
@@ -44,6 +47,11 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HOMEBREW_TAP_TOKEN: ${{ steps.app-token.outputs.token }}
+          QUILL_SIGN_P12: ${{ secrets.QUILL_SIGN_P12 }}
+          QUILL_SIGN_PASSWORD: ${{ secrets.QUILL_SIGN_PASSWORD }}
+          QUILL_NOTARY_KEY: ${{ secrets.QUILL_NOTARY_KEY }}
+          QUILL_NOTARY_KEY_ID: ${{ secrets.QUILL_NOTARY_KEY_ID }}
+          QUILL_NOTARY_ISSUER: ${{ secrets.QUILL_NOTARY_ISSUER }}
 
   # Linear release tracking. Only runs for tag pushes, not manual dispatch.
   linear-sync:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,9 +14,21 @@ jobs:
     name: release
     runs-on: ubuntu-latest
     concurrency: release-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


lukepiette (approved):
Approving. Verified the pipeline end to end rather than just the diff: the quill hook runs post-lipo and pre-archive, so the release checksums, brew formula, and cask sha256 are all computed from the signed artifact (no checksum-mismatch bug); a signing or notary failure fails the whole run before anything publishes; secrets are env-scoped to the goreleaser step, never on argv, and unreachable from fork PRs (no `pull_request_target` in the repo, and all five secrets already exist in repo settings); UPX stays scoped to linux so it cannot rewrite the signed Mach-O; not stapling is correct since bare Mach-O binaries cannot be stapled and quill zips for submission internally.

Comments, in priority order (none blocking):

1. **Harden the quill install.** `curl https://get.anchore.io/quill | sudo sh` pipes an unverified installer to root in the same job that later receives the Developer ID `.p12` and notary key in env. The quill version is pinned but the installer is not; a compromised installer is a direct exfiltration path for the signing identity. Fetch the quill release tarball from GitHub and verify a sha256 pinned in the workflow instead.
2. **The timeout comment is inverted.** quill v0.7.1 bounds its own notary wait: `DefaultStatus()` is `Wait: true, PollSeconds: 10, TimeoutSeconds: 900`, and `PollStatus` wraps the poll loop in `context.WithTimeout` — a stalled submission fails at ~15 min on its own, so the 6h-lock scenario the comment fears cannot happen. The actual risk is the opposite: first-time Apple submissions routinely exceed 15 minutes, which would fail a release Apple would have accepted. Keep `timeout-minutes: 45` as belt and braces, fix the comment, and consider a committed `.quill.yaml` with `status: { timeout-seconds: 2400 }` (or document that re-running is cheap — resubmission of an already-scanned hash is fast). Watch the first tag; it is the live notary test.
3. Minor: the `legacy` per-arch darwin binaries still ship unsigned (disclosed, only pre-change self-updaters fetch them); `goreleaser-action` `version: latest` is now part of the signing chain and worth pinning in the same spirit as the quill pin; local `goreleaser snapshot` now requires quill on the developer's machine (ad-hoc mode, no credentials) — worth a README line.

PR #324, opened 2026-08-03
fix(exec): return the error so a failed run exits non-zero

justinwlin:
## what

`runpodctl exec python` printed its error and **exited 0**. A caller chaining on success ran anyway:

```bash
runpodctl exec python train.py && ./deploy.sh   # deployed on a failed run
```

The handler used `Run`, which has no way to report failure, so the error never reached the `Execute` sink and carried no `code` either.

## how

`Run` → `RunE`, return the wrapped error. Nothing else needed changing — `PodSSHConnection` and `PythonOverSSH` already return errors instead of exiting.

Progress output stays on **stdout** and the 5-minute `maxPollTime` wait is unchanged: AGENTS.md requires legacy commands preserve their stdout exactly, and that output is progress, not an error.

| before | after |
|---|---|
| error on stderr, plaintext, **exit 0** | flat json with a `code`, **exit 1** |
| stdout progress | unchanged |
| deprecation notice on stderr | unchanged |

<details>
<summary>e2e, per AGENTS.md</summary>

Throwaway `HOME` with no ssh key, so no resources were created and nothing needed cleanup:

```
$ runpodctl exec python /tmp/nope.py
exit 1
stdout: Running remote Python shell...
stderr: warning: 'runpodctl exec' is deprecated; use 'runpodctl ssh info <pod-id>' …
        note: legacy exec behavior is kept for backward compatibility
        {"error":"executing python over ssh: getting SSH connection: reading private
          SSH key from …/.runpod/ssh/runpodctl-ssh-key: no such file or directory",
         "code":"cli_error"}
```

`go build ./...`, `go vet ./...`, `go test ./...` clean. The new test in `cmd/exec/commands_test.go` drives the real error path with no network and no 5-minute wait — an empty `HOME` fails the ssh key read, which happens before polling starts.

</details>

## notes

- Stacked on #319, which corrected the README's description of this bug. Merge that first; this PR moves `exec` out of the known-gaps table and marks it converted in AGENTS.md.
- Found while reviewing #319. The README used to attribute this to CON-816, but that ticket only ever covered `project` and is now closed as a duplicate of CON-874 — so this one is untracked. Happy to file a ticket if you want it linked.


justinwlin:
Folded into #319 as b4f5d18 — one PR is easier to review than a two-deep stack, and the readme paragraph this touches was already being rewritten there.

PR #329, opened 2026-08-19
feat(registry): read the registry password from stdin

### LINKED ISSUE #327: runpodctl registry create accepts a registry password only as a command-line flag

justinwlin:
Closes #327

## Context

`registry create` stores a container-registry credential so Runpod can pull private images. It needs a username and a password or access token:

```console
$ runpodctl registry create --name ghcr --username me --password <secret>
```

## The issue

`--password` was the only way to supply the secret, and cobra marked it **required** — so there was no way to run the command without putting a long-lived credential in `argv`, where it is readable from the process table and written to shell history. No `--password-stdin`, no file, no prompt.

The reporter did not use the command at all: they created the registry auth in the console and read back only the id with `registry list`, specifically to avoid the flag. So the CLI was unusable for the one workflow it exists to serve, by anyone who treats registry tokens as secrets.

## The fix

- **`--password-stdin`**, matching `docker login`.
- **A no-echo prompt** when stdin is a terminal and neither flag is given.
- **`--password` is unchanged** and stays a first-class option — no warning, no deprecation. Whether argv exposure matters is the caller's call.
- The two password flags are mutually exclusive.

```console
$ echo "$TOKEN" | runpodctl registry create --name ghcr --username me --password-stdin
$ runpodctl registry create --name ghcr --username me --password-stdin < token.txt
$ runpodctl registry create --name ghcr --username me      # prompts, no echo
```

## Two things I'd like a second opinion on

**1. I removed `MarkFlagRequired("password")`.** Cobra enforces required flags *before* `RunE`, so leaving it would reject the stdin and prompt paths before they could supply a value — the new features cannot work with that line present. The guarantee therefore moves from the framework into `resolvePassword`, which is hand-written and so can be wrong in a way the framework could not. I traced every branch and `registry_test.go` now pins the required/not-required split per flag, but "I checked my own logic" is the claim worth verifying rather than accepting.

**2. I added four lines to `cmd/root.go`, outside the registry package.** This is the repo's first `MarkFlagsMutuallyExclusive` site, so cobra's `ValidateFlagGroups` message reached the central error classifier unrecognised and reported `cli_error` ("something broke") instead of `usage_error` ("you used it wrong"). Adding it to `usageErrorPrefixes` fixes the label.

Two caveats. That file classifies errors for the whole CLI, so it is the only part of this PR that can affect commands I did not touch. And it works by matching cobra's English message text — five entries already do this, with a comment warning to re-check them on cobra upgrades, but this adds a sixth hostage to that fragility. Additive and narrow, but worth agreeing to rather than inheriting.

`golang.org/x/term` moves from indirect to direct. No new dependency.

## Two defects I found and fixed in self-review

**Multi-line stdin was rejected.** I had guarded on the assumption that a credential never spans lines. A gcr.io / Artifact Registry credential is an entire service-account json key passed as the password with username `_json_key` — the same shape `docker login --password-stdin` takes. I confirmed against the live api that it accepts a multi-line password, so the guard rejected valid input and broke the exact workflow `--password-stdin` exists for. Removed, with a regression test.

**Ctrl-C at the prompt left the terminal broken.** `term.ReadPassword` disables echo and restores it from a deferred ioctl, which SIGINT's default disposition never reaches — the process died with echo still off, leaving the caller's shell typing blind until `stty sane`. Reproduced on a pty (`ECHO` still false after SIGINT). The prompt now restores saved terminal state from a signal handler and exits 130. Re-verified on a pty: echo returns to on, exit code is 130, normal path unaffected.

## Testing

10 new tests in `cmd/registry/create_password_test.go` (13 top-level in the package, 9 subtests). `resolvePassword` takes an `io.Reader`/`io.Writer`, and the tty detector and prompt are package vars, so the whole matrix runs without a terminal: a trailing `\n`/`\r\n` is stripped while leading, inner and pre-newline whitespace survive; empty stdin rejected while a multi-line credential is preserved; a broken pipe stays uncoded (environment failure, not usage); `--password` asserted to print nothing and leave stdin unconsumed; no-source-no-tty fails fast rather than blocking on a read that never arrives.

<details>
<summary>Live e2e against the production api, resources cleaned up</summary>

Per AGENTS.md, built the binary and ran every path against the live api.

| path | result |
|---|---|
| `--password-stdin` via pipe | created ok |
| `--password` flag | created ok, stdout parseable json, stderr empty |
| `--password-stdin < file` | created ok |
| interactive prompt on a real pty | created ok, prompt shown, typed secret absent from the pty transcript (echo genuinely off) |
| multi-line `_json_key` json on stdin | created ok — the regression the self-review fix was for |

Error paths return before any api call, because `resolvePassword` runs first. Each returns `usage_error` and exits 1:

- both `--password` and `--password-stdin` together
- empty stdin with `--password-stdin`
- no password source with stdin not a terminal

`--name`/`--username` remain required. Multi-line stdin is **accepted**, not an error.

Every auth created was deleted afterwards, and `registry list` confirmed clean.

</details>

---

## Related PRs

Two independent groups. Nothing in one blocks anything in the other.

### Group A — the Windows install command (runpod/runpodctl#311)

One broken command that had been copy-pasted into three repos. Each PR fixes its own copy; **any order, no dependencies.**

| PR | Repo | Change |
|---|---|---|
| runpod/runpodctl#328 | runpodctl | `README.md` — fix the command, and install to `PATH` |
| runpod/docs#805 | docs | `runpodctl/overview.mdx` — same fix on docs.runpod.io |
| runpod/runpod-plugins-official#49 | plugins | delete `reference/install.md`, the third copy |

### Group B — secure registry passwords (runpod/runpodctl#327)

**Order matters.** #51 documents a flag that does not exist on `main` until #329 merges, so it stays draft until then.

| # | PR | Repo | Change |
|---|---|---|---|
| 1 | runpod/runpodctl#329 ← **this PR** | runpodctl | `--password-stdin`, no-echo prompt, mutually exclusive flags |
| 2 | runpod/runpod-plugins-official#51 | plugins | skill guidance + regression eval (**draft** until #329 lands) |

The only thing the two groups share is that both touched the runpodctl skill, which is what prompted Group A's deletion: the skill had been keeping its own copy of install instructions the runpodctl README already owned.

### DIFF
--- a/cmd/registry/create.go
+++ b/cmd/registry/create.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 
 	"github.com/runpod/runpodctl/internal/api"
 	"github.com/runpod/runpodctl/internal/clierr"
@@ -58,8 +60,36 @@ func init() {
 // promptPassword is a package variable so tests can exercise the prompt branch
 // without a controlling terminal.
 var promptPassword = func(stderr io.Writer) (string, error) {
+	fd := int(os.Stdin.Fd())
+
+	// ReadPassword disables echo and restores it from a deferred ioctl, which a
+	// signal's default disposition never reaches: a ctrl-c at the prompt would
+	// kill the process with echo still off and leave the caller's shell typing
+	// blind until they ran `stty sane`. Restore it from a handler instead.
+	state, err := term.GetState(fd)
+	if err != nil {
+		return "", fmt.Errorf("failed to read the terminal state: %w", err)
+	}
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
+	defer signal.Stop(sigCh)
+	done := make(chan struct{})
+	defer close(done)
+	go func() {
+		select {
+		case <-sigCh:
+			_ = term.Restore(fd, state)
+			fmt.Fprintln(stderr)
+			// 128 + SIGINT, the shell convention for "died on a signal". The
+			// deferred restores below are deliberately skipped: the terminal is
+			// already back and there is nothing else to unwind at a prompt.
+			os.Exit(130)
+		case <-done:
+		}
+	}()
+
 	fmt.Fprint(stderr, "registry password: ")
-	secret, err := term.ReadPassword(int(os.Stdin.Fd()))
+	secret, err := term.ReadPassword(fd)
 	// ReadPassword swallows the user's newline, so the next line of output would
 	// otherwise start on the prompt line.
 	fmt.Fprintln(stderr)
@@ -86,17 +116,14 @@ func resolvePassword(stdin io.Reader, stderr io.Writer, flagValue string, fromSt
 		if err != nil {
 			return "", fmt.Errorf("failed to read the password from stdin: %w", err)
 		}
-		// strip only the trailing line ending a pipe or heredoc adds. leading and
-		// inner whitespace can be part of the credential and must survive.
+		// strip only the trailing line ending a pipe or heredoc adds. everything
+		// else survives: a gcr.io / artifact registry credential is a whole
+		// service-account json key (username _json_key), so the value is
+		// legitimately multi-line, and the api accepts it. leading and inner
+		// whitespace can be significant too.
 		secret := strings.TrimSuffix(strings.TrimSuffix(string(data), "\n"), "\r")
 		if secret == "" {
 			return "", clierr.Usagef("the password read from stdin is empty")
-		}
-		// a token never spans lines, so this is a redirected file with extra
-		// content rather than a credential. the api would reject it with a less
-		// obvious error.
-		if strings.ContainsAny(secret, "\r\n") {
-			return "", clierr.Usagef("the password read from stdin spans multiple lines")
 		}
 		return secret, nil
 
--- a/cmd/registry/create.go
+++ b/cmd/registry/create.go
@@ -25,7 +25,7 @@ the password can come from --password, from stdin with --password-stdin, or from
 an interactive prompt when neither flag is given. --password-stdin keeps the
 credential out of the process table and your shell history.`,
 	Example: `  # read the password from a pipe
-  echo "$REGISTRY_TOKEN" | runpodctl registry create --name ghcr --username me --password-stdin
+  printenv REGISTRY_TOKEN | runpodctl registry create --name ghcr --username me --password-stdin
 
   # read it from a file without exposing it to the process table
   runpodctl registry create --name ghcr --username me --password-stdin < token.txt
@@ -71,7 +71,7 @@ var promptPassword = func(stderr io.Writer) (string, error) {
 		return "", fmt.Errorf("failed to read the terminal state: %w", err)
 	}
 	sigCh := make(chan os.Signal, 1)
-	signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
+	signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM, syscall.SIGQUIT)
 	defer signal.Stop(sigCh)
 	done := make(chan struct{})
 	defer close(done)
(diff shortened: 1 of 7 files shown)


cmd/registry/create.go
```diff
@@ -2,38 +2,165 @@ package registry
...
-	Long:  "create a new container registry authentication",
-	Args:  cobra.NoArgs,
-	RunE:  runCreate,
+	Long: `create a new container registry authentication
+
+the password can come from --password, from stdin with --password-stdin, or from
+an interactive prompt when neither flag is given. --password-stdin keeps the
+credential out of the process table and your shell history.`,
+	Example: `  # read the password from a pipe
+  echo "$REGISTRY_TOKEN" | runpodctl registry create --name ghcr --username me --password-stdin
```
lukepiette: Could we use `printenv REGISTRY_TOKEN` here, matching the companion plugin PR? `echo` treats some valid credential values as options—for example, a token equal to `-n` produces zero bytes in both bash and zsh—so runpodctl receives an empty password. `printenv` puts only the environment-variable name in the producer's argv and avoids shell-specific `echo` behavior.

cmd/registry/create.go
```diff
@@ -2,38 +2,165 @@ package registry
...
+	// ReadPassword disables echo and restores it from a deferred ioctl, which a
+	// signal's default disposition never reaches: a ctrl-c at the prompt would
+	// kill the process with echo still off and leave the caller's shell typing
+	// blind until they ran `stty sane`. Restore it from a handler instead.
+	state, err := term.GetState(fd)
+	if err != nil {
+		return "", fmt.Errorf("failed to read the terminal state: %w", err)
+	}
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
```
lukepiette: Please include `syscall.SIGQUIT` in the terminal-restoration handler and add the expected exit-code-131 test. On Unix, Ctrl-\ sends SIGQUIT; the default Go handling terminates without running `ReadPassword`'s deferred restore, so terminal echo can remain disabled—the same failure this block prevents for Ctrl-C.

cmd/registry/create.go
```diff
@@ -2,38 +2,165 @@ package registry
...
+
+	default:
+		// non-interactive with no password source is a scripting mistake, and
+		// silently prompting into a pipe would hang instead of reporting it.
+		return "", clierr.Usagef("a password is required; pass --password-stdin to read it from stdin, or --password")
+	}
 }
 
 func runCreate(cmd *cobra.Command, args []string) error {
+	password, err := resolvePassword(cmd.InOrStdin(), cmd.ErrOrStderr(), createPassword, createPasswordStdin)
```
lukepiette: Non-blocking: consider constructing `api.NewClient()` before reading or prompting for the registry password. That lets a missing Runpod API key fail immediately instead of consuming a one-shot stdin secret or asking the user to enter a password that cannot be submitted.

lukepiette (approved):
Approving. The inline comments are non-blocking follow-up improvements; the core implementation is sound.

PR #338, opened 2026-09-09
STO-67: fast-fail Model Repo storage quota check

jebenexer:
Resolves [STO-67](https://linear.app/runpod/issue/STO-67/create-model-repo-storage-limit).
Server-side counterpart: runpod/RunPod#5929.

#339 (STO-366, pre-signed URL batching) is independent — this check runs before
session creation either way.

## Problem

`runpodctl` already knows the total model size before uploading, but goes
straight into session creation with no check against the account's storage
quota. An over-quota upload should fail immediately with a useful message,
not after creating a model and partial upload sessions server-side.

## Changes

- `api.GetModelRepoStorageUsage` wraps the new `modelRepoStorageUsage` query.
- `checkModelRepoStorageQuota` compares the total upload size against
  `availableBytes` and returns an actionable error when it would be exceeded.
  It **fails open**: a query error or unparseable response warns and proceeds,
  since a client-side pre-check must never itself break a valid upload. Skipped
  when enforcement is off (`Enforced: false`) or no quota is set
  (`AvailableBytes: nil`). `createModelRepoUpload`'s server-side gate is still
  the real limit.
- Wired into `runAddModel` after file-name/file-size validation and before
  `addModelToRepo`, covering both `--model-path` and single-file flows.

## Testing

- [x] `go vet ./...`, `go build ./...`, `go test ./...`, `govulncheck ./...`
- `checkModelRepoStorageQuota` allow/deny/boundary/fail-open, and
  `modelUploadRequestedBytes`' directory-vs-flag size resolution.
- API-level tests for `GetModelRepoStorageUsage`, including its entry in the
  STO-357 typed-error table.
- End-to-end test confirming `addModelToRepo` and `createModelRepoUpload` are
  never reached when the check rejects the upload.

---

> [!NOTE]
> **Low Risk**
> Client-only pre-check that fails open on API errors; upload enforcement still happens server-side, with broad test coverage and no auth or data-model changes.
> 
> **Overview**
> Adds a **client-side Model Repo storage quota pre-check** on `runpodctl model add` so over-quota uploads fail before any server-side model or upload session is created.
> 
[...]

PR #339, opened 2026-09-09
STO-366: runpodctl model upload pre-signed URL batching

jebenexer:
Resolves [STO-366](https://linear.app/runpod/issue/STO-366/runpodctl-model-upload-pre-signed-url-batching).
Server-side counterpart: runpod/RunPod#5931.

Independent of #338 (STO-67), which touches the quota fast-fail check — that
runs before upload-session creation either way.

## Problem

`runpodctl` uploads a model one file at a time: initiate file A, wait for
GraphQL to return A's pre-signed URL, upload A, repeat. For models with many
small files, most of the wall clock goes to those per-file round trips rather
than to the byte transfer.

## Changes

- `api.CreateModelRepoUploadBatch` wraps the new `createModelRepoUploadBatch`
  mutation, which creates upload sessions for a whole file manifest at once.
- `uploadModelFiles` chunks the manifest into batches of up to
  `modelRepoUploadBatchSize` (500) and calls the batch mutation once per chunk.
  Chunks after the first are pinned to the model version the first chunk
  established, so a large manifest still lands on a single version.
- Byte upload and progress tracking (`completeModelUploadFile`) are unchanged —
  only session creation moves.

## Testing

- [x] `go vet ./...`, `go build ./...`, `go test ./...`, `govulncheck ./...`
- Existing `uploadModelFiles` and directory-upload tests now stub the batch
  mutation instead of the per-file one, with the same assertions plus
  one-batch-per-manifest.
- New test for a manifest larger than `modelRepoUploadBatchSize` splitting
  across multiple batch calls.

---

> [!NOTE]
> **Medium Risk**
> Changes the core directory-upload path and depends on a new server mutation; incorrect batching or version pinning could split uploads across versions or fail large manifests.
> 
> **Overview**
> Adds **`CreateModelRepoUploadBatch`** in the API layer, calling the new `createModelRepoUploadBatch` GraphQL mutation so many files can get presigned upload sessions in one round trip instead of one HTTP call per file.
> 
> **`runpodctl model add --model-path`** now uses that batch API in **`uploadModelFiles`**: manifests are sent in chunks of up to **500** files (`modelRepoUploadBatchSize`, aligned with the server cap). The first batch establishes the model version UUID; later chunks pass **`modelVersionUuid`** so everything still attaches to a single version. Multipart byte upload, progress, and per-session **`completeModelRepoUpload`** behavior are unchanged; only session creation is batched. The explicit single-file upload flags still go through **`CreateModelRepoUpload`**.
> 
> Tests cover one-request batching, empty/mismatched responses, GraphQL errors, multi-batch manifests, and access-denied handling for the new API entry point.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c277b4ca58be9326abe094fa26fd8999c0a6668. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


api/model.go
```diff
@@ -779,6 +809,162 @@ func CreateModelRepoUpload(input *CreateModelRepoUploadInput) (*ModelRepoMutatio
...
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	rawData, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("statuscode %d: %s", res.StatusCode, string(rawData))
```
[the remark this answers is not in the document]
jebenexer: Confirmed and fixed in 352f56f — these were the only two sites in `api/model.go` still using bare errors where every sibling mutation wraps them.

Both now use the same helpers as `CreateModelRepoUpload`: `modelRepoHTTPError` on non-200 and `modelRepoGraphQLError` on top-level `errors`. The remaining bare errors in the function (`data is nil`, `!result.Success`, upload-count mismatch) match the sibling, which leaves those untyped too.

Also added `CreateModelRepoUploadBatch` to the `TestModelRepoFunctionsReturnGraphQLErrorOnAccessDenied` table in `api/model_repo_access_test.go` — that table is the designated home for the STO-357 contract and is what should have caught this. Verified it fails on the pre-fix code and passes after.

jebenexer:
@cursor review

api/model.go
```diff
@@ -247,6 +247,36 @@ type CreateModelRepoUploadInput struct {
 	CredentialReference string                 `json:"credentialReference,omitempty"`
 }
 
+// ModelRepoUploadBatchFileInput is one file of a createModelRepoUploadBatch manifest,
+// mirroring CreateModelRepoUploadInput's per-file fields.
+type ModelRepoUploadBatchFileInput struct {
+	FileName      string `json:"fileName"`
+	FileSizeBytes string `json:"fileSizeBytes"`
+	PartSizeBytes string `json:"partSizeBytes,omitempty"`
```
taylorburmeister-runpod, line 255: What is `PartSizeBytes`?

api/model.go
```diff
@@ -247,6 +247,36 @@ type CreateModelRepoUploadInput struct {
 	CredentialReference string                 `json:"credentialReference,omitempty"`
 }
 
+// ModelRepoUploadBatchFileInput is one file of a createModelRepoUploadBatch manifest,
+// mirroring CreateModelRepoUploadInput's per-file fields.
+type ModelRepoUploadBatchFileInput struct {
+	FileName      string `json:"fileName"`
+	FileSizeBytes string `json:"fileSizeBytes"`
```
taylorburmeister-runpod, line 254: Seems strange to me that this is a string and not an int.

api/model.go
```diff
@@ -779,6 +809,162 @@ func CreateModelRepoUpload(input *CreateModelRepoUploadInput) (*ModelRepoMutatio
...
+
+	name := strings.TrimSpace(input.Name)
+	if name == "" {
+		return nil, fmt.Errorf("name cannot be empty")
+	}
+	if len(input.Files) == 0 {
+		return nil, fmt.Errorf("files cannot be empty")
+	}
+
+	files := make([]map[string]interface{}, 0, len(input.Files))
```
taylorburmeister-runpod, line 827: All of the untyped `map[string]interface`s being used here feel strange to me. I gather this is the expected input to GraphQL, but is there not some way to pass it typed structs and have it use reflection?

cmd/model/addModelToRepo.go
taylorburmeister-runpod, line 1: Nit, and predates this PR, but the camelcase name of this file does not follow conventions.

taylorburmeister-runpod:
I left a few comments and also asked an agent to review with a focus on correctness, performance, and clarity, in that order, and it gave the following feedback, which seems legitimate:

Correctness

**1. Uploads are paired to files by slice position, with no verification of ordering (cmd/model/addModelToRepo.go:924-925)**
The batch loop does upload := result.Uploads[i] and applies it to chunk[i], i.e. it assumes the server returns uploads in the exact order of the request files. The ModelRepoUpload payload carries key/keyPrefix/sessionId but no fileName, so there is nothing to correlate a returned session back to its input file. The length guard in api.CreateModelRepoUploadBatch (api/model.go, len(result.Uploads) != len(input.Files)) only checks the count, not the correspondence. If the server ever returns sessions in a different order, completeModelUploadFile uploads file A's bytes against file B's presigned URL — silent cross-contamination that no error surfaces.
- Fix: confirm the server's schema guarantees response order matches request order, and add a comment documenting that dependency; ideally have the mutation echo fileName per upload and pair by name instead of index.

**2. Presigned URLs for a whole 500-file chunk are minted up front but consumed sequentially (cmd/model/addModelToRepo.go:908-943)**
Previously each presigned URL was created immediately before its bytes were uploaded. Now up to modelRepoUploadBatchSize (500) URLs are created in one call, then files are uploaded one at a time. Each ModelRepoUpload has ExpiresInSeconds; the URL for file #500 is issued at batch time but not used until 499 files have transferred. For large models / slow links this can exceed the expiry and fail late in the batch. Worth verifying expiresInSeconds comfortably exceeds worst-case sequential upload time for a full chunk, or lowering the batch size / handling expiry with a re-batch.

Performance

**3. Byte uploads remain fully sequential (cmd/model/addModelToRepo.go:924-943)**
The batching removes the per-file session-creation round trips (the stated goal), but the actual transfers still run one file at a time. For the motivating case — "models with many small files" — the dominant remaining cost is now per-file S3 request latency, which a bounded worker pool over the batch's uploads would cut substantially. Not a regression, but the batch structure now makes this the obvious next win. (Note: parallel completion would interact with finding #2 — parallelism would also shorten the expiry exposure window.)

Repeated code

**4. The upload-session GraphQL selection set is duplicated (api/model.go CreateModelRepoUpload vs CreateModelRepoUploadBatch)**
The uploads { sessionId … parts { … } completeUrl abortUrl } and model { … } / version { … } selection blocks are copy-pasted between the two mutations (this is also why the diff shows a whitespace-only reformat of the original block). Extracting these into a shared GraphQL fragment constant would keep the two in sync as fields are added.

Clarity (minor)

**5. Loop-scoped invariant check reads slightly obscurely (cmd/model/addModelToRepo.go:915-922)**
if modelVersionUUID == "" && end < len(files) correctly means "only error about a missing version UUID if more batches will follow." That's the intended port of the old i < len(files)-1, but the intent isn't obvious at the call site — a one-line comment ("later batches must pin to this UUID; the final batch may legitimately return none") would help.

api/model.go
jebenexer: Added a doc comment -- it's the optional multipart chunk size in bytes, matching `CreateModelRepoUploadInput.PartSizeBytes` and the CLI's `--part-size` flag. Server picks a default when omitted.

api/model.go
jebenexer: Mirrors `CreateModelRepoUploadInput.FileSizeBytes`, which is a string because the server's GraphQL field is typed `String`, not `Int` -- GraphQL's `Int` is 32-bit and can't hold file/part sizes above ~2GiB (and JSON numbers above 2^53 lose precision). Added a comment on the field explaining this.

api/model.go
jebenexer: This mirrors the pattern already used by every other mutation in this file (`AddModelToRepo`, `CreateModelRepoUpload`, etc.) -- the manual map + `addString` trim-then-omit helper predates this PR. Typed structs already exist (`CreateModelRepoUploadBatchInput` has proper `omitempty` tags) and marshaling them directly would work for most fields, but the codebase consistently rebuilds a map instead, likely so empty-after-trim strings get omitted too (bare `omitempty` only catches `""`, not whitespace). Would rather change this consistently across the whole file than diverge just the new function from its siblings.

cmd/model/add_model_to_repo.go
jebenexer: Renamed all four camelCase files in `cmd/model` (and their test files) to snake_case: `addModelToRepo.go` -> `add_model_to_repo.go`, `getModels.go` -> `get_models.go`, `removeModel.go` -> `remove_model.go`, plus `addModelDeleteAfterUpload_test.go` and the two `*_test.go` siblings. Full directory consistency now.

jebenexer:
Addressed all of these:

- **5/6 (dedup + clarity comment)**: `CreateModelRepoUpload` and `CreateModelRepoUploadBatch`'s GraphQL selection sets now share `modelRepoUploadSessionFields` / `modelRepoModelFields` / `modelRepoVersionFields` constants in `api/model.go` instead of being copy-pasted. Added a one-line comment on the `modelVersionUUID == "" && end < len(files)` guard explaining that only the final chunk may legitimately omit a version uuid.

- **1 (ordering)**: confirmed against the server resolver (`RunPod#5977` / `modelRepo.ts`) -- `plannedFiles`/`fileWork`/`sessions`/`uploads` are all built by iterating the same ordered input `files` array via one shared index throughout, so index-based pairing between `result.Uploads[i]` and `chunk[i]` is correct today. Documented that as an explicit contract on `uploadModelFileChunk` in `cmd/model/add_model_to_repo.go`, since `ModelRepoUpload` carries no `fileName` to re-verify against. I'll be filing a ticket for the server-side `fileName` echo to enable name-based reconciliation instead of positional.

- **2/4 (expiry risk / sequential uploads)**: reworked both loops that were serial per-file round trips:
  - `uploadModelFileChunk` now uploads up to `modelRepoUploadConcurrency` (4) files per chunk concurrently instead of one at a time, cutting the presigned-URL expiry exposure window and wall-clock for many-small-file manifests.
  - The finalize step no longer calls `completeModelRepoUpload` once per session at all: it now sends the entire uploaded-session list -- as it exists once every chunk has finished uploading -- as a single `completeModelRepoUploadBatch` request, using one aliased `completeModelRepoUpload` field per session in one GraphQL document/HTTP round trip. Note: since `completeModelRepoUpload`'s return type is non-null, a single session erroring nulls that whole response per the GraphQL spec, so on any one failure the batch's overall outcome is reported as unconfirmed rather than assuming the rest succeeded or failed; documented on `CompleteModelRepoUploadBatch`.

Also renamed `addModelToRepo.go`/`getModels.go`/`removeModel.go` and their test files to snake_case per the filename nit below.

taylorburmeister-runpod (approved):

PR #340, opened 2026-09-09
STO-691: fix `serverless update --model-reference`/`--clear-models` on GPU endpoints

jebenexer:
Fixes [STO-691](https://linear.app/runpod/issue/STO-691).

## Problem

`serverless update <id> --model-reference <ref>` and `--clear-models` both
failed on every GPU endpoint:

```
{"error":"failed to update model references: graphql error: gpuId(s) is required for a gpu endpoint","code":"graphql_error"}
```

## Root cause

`UpdateEndpointModels` round-trips the endpoint's config through `saveEndpoint`,
reading the gpu selection from `endpoint.GpuIDs`. REST `GET /v1/endpoints/{id}`
never returns `gpuIds` — only `gpuTypeIds`, gpu *type* names (`"NVIDIA A40"`) in
a different identifier space from what `saveEndpoint` wants (`"AMPERE_48"`). So
`GpuIDs` was always empty, and `saveEndpoint` rejects an empty `gpuIds` on any
endpoint without `instanceIds`.

## Fix

Read `gpuIds` over GraphQL (`myself { endpoint(id:) { gpuIds } }`, the query
`serverless model-status` already uses) and pass it to `saveEndpoint` untouched.

Reconstructing it from `gpuTypeIds` would be lossy, so this deliberately
doesn't: `gpuIds` carries `-` exclusions restricting an endpoint to a subset of
a pool (`AMPERE_48,-NVIDIA RTX A6000`) that type names can't express, so a
derived pool id silently widens the endpoint's permitted hardware — caught in
review by @justinwlin. An empty `gpuIds` with no `instanceIds` now errors
instead of writing away the gpu selection.

Also hardens `ResolveServerlessGpuPoolID`, which returned its input unchanged
with a nil error when the pools query failed, letting a gpu type name reach
`saveEndpoint` and get rejected as `Invalid GPU Pool ID` — masking the real
failure. It now propagates, which also stops `serverless create --gpu-id` when
the pools query is down.

## Testing

Unit: an exclusion survives the write verbatim; a failed or empty `gpuIds` read
errors without calling `saveEndpoint`; a cpu endpoint still updates; a failed
pools query propagates. All four fail against the previous commit.

Live on **production**, endpoint restricted to one gpu of the two-gpu
`AMPERE_48` pool (workers 0/0, no requests, deleted after):

| step | graphql `gpuIds` | rest `gpuTypeIds` |
| --- | --- | --- |
| baseline (A40-only) | `AMPERE_48,-NVIDIA RTX A6000` | `[NVIDIA A40]` |
| previous commit, `--clear-models` | `AMPERE_48` (widened) | `[NVIDIA A40, NVIDIA RTX A6000]` |
| this commit, both flags | `AMPERE_48,-NVIDIA RTX A6000` | `[NVIDIA A40]` |

Fault injection (proxy 503s one query, REST + `saveEndpoint` still prod): 503 on
`serverlessGpuPools` leaves the update unaffected — this path no longer queries
pools; 503 on the `gpuIds` read exits 1 naming that failure, with
`saveEndpoint` never called.

- [x] `go vet ./...`, `go build ./...`, `go test ./...`, `govulncheck ./...`
- [x] `go test -tags e2e ./e2e` — `TestCLI_ServerlessUpdateModelReferenceOnGPUEndpoint` against production

The e2e test added in this PR's first revision was committed unrun and could
never pass: it asserted `modelReferences` on `update`'s output, which comes
from a REST read that omits the field. It now reads them via `serverless
model-status`, and reads the gpu selection from the `gpuIds` key the cli
renames `gpuTypeIds` to. `AGENTS.md` documents both traps.

---

> [!NOTE]
> **Medium Risk**
> Touches serverless endpoint updates that full-replace config via saveEndpoint; scope is narrow (GPU id resolution on model-reference/clear-models) but mistakes could mis-assign GPU pools or break updates.
> 
> **Overview**
> Fixes **STO-691**: `serverless update --model-reference` and `--clear-models` no longer fail on GPU endpoints with `gpuId(s) is required for a gpu endpoint`.
> 
> Those updates round-trip the endpoint through GraphQL `saveEndpoint` after a REST `GET`. REST only returns **`gpuTypeIds`** (display names), not **`gpuIds`** (pool ids like `ADA_24`), so `gpuIds` was always empty on write. **`UpdateEndpointModels`** now maps `gpuTypeIds` → pool id via **`ResolveServerlessGpuPoolID`** (same as create) when `gpuIds` is missing; unknown types error instead of sending an empty `gpuIds`. The **`Endpoint`** struct gains **`GpuTypeIDs`** for the REST read shape, and **AGENTS.md** documents the asymmetry.
> 
> Unit tests cover successful resolution and unresolvable GPU types; a new e2e test creates a GPU endpoint and exercises both update flags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8e04d4d0f7e1e9c5fa3de986bf71ababc1eb22a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jebenexer:
@cursor review

justinwlin:
Live-tested this PR at `c8e04d4d0f7e1e9c5fa3de986bf71ababc1eb22a` against production. The ordinary GPU model-update path now succeeds, but I found one configuration-preservation bug and one error-handling issue.

### 1. Preserve GPU exclusions when updating model references (P1)

The new conversion in `internal/api/endpoints.go:295–301` maps REST GPU type IDs to whole pools, dropping exclusions that restrict an endpoint to particular GPUs within a pool.

Live reproduction:

1. Create a temporary GPU endpoint with `workersMin=0` and `workersMax=0`.
2. Restrict it through REST PATCH to `{"gpuTypeIds":["NVIDIA A40"]}`.
3. Verify GraphQL reports `gpuIds: "AMPERE_48,-NVIDIA RTX A6000"` and REST reports only `NVIDIA A40`.
4. Run this PR's binary: `runpodctl serverless update <id> --clear-models`.
5. The command succeeds, but GraphQL now reports `gpuIds: "AMPERE_48"`, and REST reports both `NVIDIA A40` and `NVIDIA RTX A6000`.
6. Reset the A40-only restriction and run `--model-reference https://huggingface.co/Qwen/Qwen2.5-0.5B-Instruct:main`. This also succeeds and removes the exclusion.

Changing model references should not expand the endpoint's permitted hardware. Prefer fetching and preserving the original GraphQL `gpuIds`, or reconstruct both pool IDs and exclusions from the REST selection. Add coverage for a subset of a multi-GPU pool; the current test includes the entire pool and cannot catch this. A local regression test using A40-only input also failed, producing `AMPERE_48` instead of `AMPERE_48,-NVIDIA RTX A6000`.

### 2. Stop when the GPU-pool lookup fails (P2)

`ResolveServerlessGpuPoolID` returns its input unchanged with a nil error when `ListServerlessGpuPools` fails. That fallback can make sense for an already-valid pool ID, but this new caller passes REST GPU type names.

Verified with targeted fault injection: a localhost proxy returned HTTP 503 only for `serverlessGpuPools`; endpoint REST reads and the subsequent `saveEndpoint` mutation still used production. The CLI continued and sent:

```json
{"gpuIds":"NVIDIA A40,NVIDIA RTX A6000"}
```

Production rejected the mutation with `Invalid GPU Pool ID: "NVIDIA A40"`. The CLI exited 1 with `graphql_error`, masking the original lookup failure. The endpoint remained unchanged in this failure case. The lookup outage was simulated; the invalid mutation and backend rejection were real.

Please propagate the lookup failure and avoid the mutation when conversion cannot be completed. Add a pool-query failure test asserting that `saveEndpoint` is not called.

### Validation and cleanup

- Existing tests passed in `internal/api`, `api`, and `cmd/serverless` on this PR head.
- Both model-update flags reproduced the GPU-selection expansion against production.
- Maximum workers remained zero throughout; no inference requests were submitted.
- All temporary endpoints and templates were deleted, and endpoint deletion was verified.

I would address the configuration-preservation issue before merging.

jebenexer:
Thanks — both fixed in `3e415f3`, and both reproduced live first so I was
verifying against the real failure rather than my reading of it.

**1. GPU exclusions (P1).** Took your first suggestion: `UpdateEndpointModels`
now reads `gpuIds` over GraphQL (`myself { endpoint(id:) { gpuIds } }`, the
query `serverless model-status` uses) and passes it through untouched. I didn't
reconstruct exclusions from the REST selection — that's inference against a
syntax this repo has never written, when the authoritative value was one query
away. Reproduced your scenario on production, A40-only on `AMPERE_48`, workers
0/0:

| step | graphql `gpuIds` | rest `gpuTypeIds` |
| --- | --- | --- |
| baseline | `AMPERE_48,-NVIDIA RTX A6000` | `[NVIDIA A40]` |
| old binary, `--clear-models` | `AMPERE_48` (widened) | `[NVIDIA A40, NVIDIA RTX A6000]` |
| new binary, both flags | `AMPERE_48,-NVIDIA RTX A6000` | `[NVIDIA A40]` |

`TestUpdateEndpointModels_PreservesGpuPoolExclusions` covers the subset case.
You were right that the old test couldn't catch this — its mock pool's members
were exactly the endpoint's two types, so there was no exclusion to lose.

**2. Pool lookup (P2).** `ResolveServerlessGpuPoolID` now propagates. I
hardened it for all callers rather than adding a strict variant, so
`serverless create --gpu-id` also stops when the pools query is down — flagging
that as a behavior change beyond this PR's scope. Repeated your fault
injection:

- *503 on `serverlessGpuPools`* — old binary swallowed it and sent
  `gpuIds: "NVIDIA A40"`, which prod rejected with exactly your
  `Invalid GPU Pool ID: "NVIDIA A40"`. New binary: unaffected, exclusion
  preserved, and the proxy log shows it never queries pools at all.
- *503 on the new `gpuIds` read* — exits 1 with `failed to read endpoint gpu
  ids: ...status 503`, `saveEndpoint` never called, endpoint untouched. That's
  your "avoid the mutation" assertion, covered by
  `TestUpdateEndpointModels_GpuIDsReadFailureSkipsWrite`.

Prod's own rejection message documents the exclusion syntax
(`To exclude specific GPU types within a pool, prefix them with...`) — nice
confirmation these are load-bearing.

**Two test bugs this shook out,** both invisible to a green local run: my first
e2e gpu assertion parsed `gpuTypeIds` from cli output, which `renameGPUKeys`
renames to `gpuIds`, so it read empty and compared `"" != ""` — vacuous. And
the e2e test in the first revision could never pass, asserting
`modelReferences` on `update`'s output when `update` renders from a REST read
that omits the field. That one's my fault for committing an unrun e2e test.

**One adjacent thing, not fixed here:** for that same reason,
`--model-reference` succeeds but never shows the refs it just set — you need
`serverless model-status`. Pre-existing and outside this feedback; happy to fix
it here or file a follow-up rather than let it linger.

Temporary endpoints deleted and deletion verified, no templates left behind.
Local suite, vet, build and govulncheck green; e2e passes against production.

justinwlin (approved):
Verified the fixes in 3e415f3 against production: both model flags preserve A40-only exclusions, updates no longer depend on the pool lookup, and a failed authoritative GPU-selection read stops before saveEndpoint.

Added 50c723c so model updates also include the server-resolved modelReferences in their output, with an explicit [] after --clear-models. This retains the existing REST endpoint fields and adds no API call.

Validation on the updated code: go test ./..., go vet ./..., go build, govulncheck (no affected code), and e2e package compilation passed. Independently live-tested both flags in JSON and YAML on an A40-only production endpoint at workers 0/0; GPU selection stayed unchanged. Rechecked the injected read-failure path. Temporary endpoint and template deleted; endpoint deletion verified. GitHub CI is still pending at submission.

No remaining blocking findings. Note that I contributed the output fix in the final commit; this approval is not an independent review of that contribution.

PR #342, opened 2026-09-15
fix(config): store the api key with private file modes (CON-1160)

OneAsterisk:
Closes [CON-1160](https://linear.app/runpod/issue/CON-1160). First of the three sub-tickets under [CON-1158](https://linear.app/runpod/issue/CON-1158); CON-1161 (ssh host keys) and CON-1159 (receive path confinement) touch disjoint files and land separately.

## the vulnerability

CWE-732. The api key is stored in plaintext in `~/.runpod/config.toml`. `initConfig` created the directory with `os.ModePerm` (0755 under a normal umask) and let viper write the file with its 0644 default, so any other user on a shared machine could read it. Every `runpodctl` invocation creates that file, so this is not limited to people who ran `config` or `doctor`.

## what changed

All of it in `initConfig` in `cmd/root.go`.

- The directory is created 0700 and `viper.SetConfigPermissions(0o600)` pins the file. That call is on the global viper, so the writes in `config`, `doctor` and `project` inherit it and those files needed no edits.
- `tightenConfigPermissions` narrows the directory, the config, and the legacy `~/.runpod.yaml` whenever they are group- or world-accessible. Without it the fix would silently no-op for every existing install, because `MkdirAll` never changes an existing directory's mode and viper only applies permissions to files it creates. The legacy file is remediated but never deleted. No-op on Windows, where chmod only toggles the read-only attribute.

Two adjacent bugs in the same function, both fixed here because a permissions change touches the same lines.

**`config --apiKey` was broken on a fresh machine.** Viper 1.19 does not update `ConfigFileUsed` after `WriteConfigAs`, and the migration branch reset the searched config name to `.runpod.yaml` unconditionally, even when no such file existed. So every later `WriteConfig()` searched for a file that was never there:

```
$ runpodctl config --apiKey=...
{"error":"error saving config: Config File \".runpod.yaml\" Not Found in ...","code":"cli_error"}
```

Naming the file up front with `SetConfigFile` pins `ReadInConfig`, `ConfigFileUsed` and every `WriteConfig` to the toml. The legacy file is now read through its own `viper.New()` instance, so a half-done migration cannot leave the global one pointed at the yaml.

**A malformed config was truncated.** The migration branch was entered on *any* `ReadInConfig` error, not just not-found, so a config the user could have repaired by hand was overwritten with defaults, taking the stored key with it:

```
$ printf 'apikey = \n' > ~/.runpod/config.toml && runpodctl version
$ cat ~/.runpod/config.toml
apikey = ''
apiurl = 'https://api.runpod.io/graphql'
```

A read error that is not "file missing" now warns on stderr and writes nothing. Commands that need a key report `no_credentials` as usual.

## verification

Unit tests in `cmd/initconfig_test.go`, five cases, one per branch. Nothing asserted any file mode anywhere in the repo before this and `initConfig` had no test at all.

The table below is measured, not inferred: I built the pre-change binary from `main` in a throwaway worktree and ran both against temp `HOME`s.

| scenario | before | after |
| --- | --- | --- |
| fresh home, directory and file | `755` / `644` | `700` / `600` |
| pre-existing wide modes | left as found | narrowed, contents byte-identical |
| `config --apiKey` on a fresh machine | fails, `Config File ".runpod.yaml" Not Found` | saves to `config.toml` |
| malformed config after `runpodctl version` | truncated to defaults | byte-identical, warning on stderr |
| `doctor` saving a key | `644` | `600` |
| legacy `~/.runpod.yaml` only | migrates, yaml left `644` | migrates, yaml left in place at `600` |
| key read from `config.toml` | `401 unauthorized` | `401 unauthorized`, unchanged |

## release note

Worth advising rotation. Keys written by earlier versions may already have been world-readable on multi-user machines, and tightening the mode now does not undo any prior exposure.

## note on the ticket

CON-1160 cites `cmd/completion.go:78` as a second site creating `~/.runpod` at 0755. It does not: that line creates `~/.config/fish/completions` for fish completion files, and 0755 is correct for it. `completion.go` is untouched here.


cmd/root.go
```diff
@@ -326,31 +330,139 @@ func Execute(ver string) {
...
+func restrictConfigPath(path string, mode os.FileMode, directory bool) error {
+	info, err := os.Lstat(path)
+	if errors.Is(err, fs.ErrNotExist) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("could not inspect %s: %w", path, err)
+	}
+	if info.Mode()&os.ModeSymlink != 0 {
+		return fmt.Errorf("config path must not be a symlink: %s", path)
```
donovanclarke: **SEVERITY** - high
**IMPACT** - Any symlinked `~/.runpod`, `config.toml`, or even an unused `~/.runpod.yaml` (stow, home-manager, a dotfile linked into `/workspace` on a pod) now fails every command with no bypass, `RUNPOD_API_KEY` included.
**SOLUTION** - Resolve with `filepath.EvalSymlinks`, then run the regular-file/dir checks and chmod on the target, or warn and skip tightening; never block on the legacy yaml when the toml loads.

donovanclarke (commented):
Verdict: COMMENT. One high finding, anchored inline. The permission work itself is the right shape (Lstat, open, SameFile, fchmod; SetConfigPermissions on the global viper) and both adjacent viper bugs are real and correctly pinned by the new tests. The high is about how far the fail-closed posture reaches.

**SEVERITY** - medium
**IMPACT** - chmod or create failures (read-only fs, foreign-owned or image-baked `~/.runpod`) now fail every command, including users authenticated only via `RUNPOD_API_KEY`; before they were tolerated.
**SOLUTION** - Warn on stderr and continue when `restrictConfigPath` chmod or `SafeWriteConfigAs` fails; continuing does not widen exposure, the file was already wide or absent.

<details open><summary><h1>Non-blocking (3)</h1></summary>

**SEVERITY** - low
**IMPACT** - A legacy `~/.runpod.yaml` that exists but fails to parse or open is skipped silently; a defaults-only toml is written and the yaml is never consulted again.
**SOLUTION** - Distinguish not-found from other errors on `legacy.ReadInConfig()` and warn on stderr, mirroring the toml default branch.

**SEVERITY** - low
**IMPACT** - Tests share the global viper and `ReadInConfig` keeps the prior map on failure, so the migration and malformed cases inherit the previous test's keys and pass by run order.
**SOLUTION** - Snapshot and restore viper state in `tempHome` (`viper.Reset()` plus re-binding the `apiKey` pflag), or assert against a fresh instance.

**SEVERITY** - low
**IMPACT** - AGENTS.md requires the runpodctl skill in runpod-plugins-official to track behavior changes; config modes, symlink rejection and fail-closed init are not reflected there.
**SOLUTION** - Open the companion skill update (config permissions section, key rotation advice) alongside this PR.

</details>

<details open>

review 1 · fable · [`bf749fc`](https://github.com/runpod/runpodctl/commit/bf749fc0b12de6fdbaf4812b9ca994424644447e)

**RAN** - none (review only)

</details>

OneAsterisk:
bugbot run

PR #343, opened 2026-09-15
fix(project): verify pod ssh host keys on the deprecated project/exec paths (CON-1161)

OneAsterisk:
Closes [CON-1161](https://linear.app/runpod/issue/CON-1161). Second of the three sub-tickets under [CON-1158](https://linear.app/runpod/issue/CON-1158). Disjoint from [#342](https://github.com/runpod/runpodctl/pull/342) (CON-1160), so the two are independent.

## the vulnerability

CWE-295. `cmd/project` disabled host key verification on both of its ssh paths: `ssh.InsecureIgnoreHostKey()` in the Go client and `StrictHostKeyChecking=no` on the openssh invocation rsync shells out to. A substituted key was accepted silently on **every** connection, not just the first, so there was no point at which a user could have noticed an interception. `runpodctl exec python` reaches the same code, so this was not limited to the `project` subcommands.

`internal/sshconnect`, behind the user-facing `ssh info`, was already fine and is untouched: it emits a plain `ssh -i <key> root@<ip> -p <port>` and leaves verification to openssh's defaults against the user's own `~/.ssh/known_hosts`.

## what changed

Both clients now share one trust store, `~/.runpod/ssh/known_hosts` (0600 inside a 0700 directory), with asymmetric roles:

- **Go client** (`podHostKeyCallback`, new `cmd/project/hostkey.go`) enrolls a pod's key on first contact, printing the fingerprint to stderr, and refuses a key that does not match an existing entry.
- **openssh/rsync** always runs after that dial, so it *requires* the pin rather than being able to create one: `StrictHostKeyChecking=yes` with `UpdateHostKeys=no`. It can neither enroll a key itself if the store went missing, nor let the server append further keys to an existing entry.

Trust is keyed on the pod, not the address: `-o HostKeyAlias=runpod-<podID>`. Runpod recycles pod ssh addresses, so an address-keyed store would report mismatches to users who did nothing wrong, and the obvious workaround, deleting the entry, trains people to ignore the one warning this change exists to raise. `HostKeyAlias` is client-side only, so there is no protocol change and the Go client looks up the same string. `CheckHostIP=no` because it defaulted to yes before openssh 8.5 and would pin the recycled ip alongside the alias.

A mismatch is reported and never repaired. The error names the pod, the offered fingerprint, the store path, and the remedy.

The callback holds an exclusive OS lock on `known_hosts.lock` (flock on unix, `LockFileEx` on windows) across read, check and append, and re-reads the store inside the lock rather than trusting the snapshot taken at callback construction. The lock file is never unlinked, since unlinking would let two processes hold locks on different inodes.

## the part worth reviewing closely

The two clients share one file, so their entry format has to agree exactly. It is asymmetric, and both halves were measured against openssh 10.3p1 and a throwaway in-process sshd rather than assumed:

| fact | why it matters |
| --- | --- |
| openssh writes a `HostKeyAlias` entry **bare, no port bracket, even on a non-default port** | byte-identical to `knownhosts.Line`, so one entry serves both |
| Go's `knownhosts` reads an unbracketed pattern as **port 22** and compares ports exactly | the callback must look up `alias:22` or the line openssh wrote will not match |
| `LogLevel=ERROR`, already on this invocation, hides the enrollment notice but **not** the refusal | which is why the Go client prints its own notice |
| `StrictHostKeyChecking=yes` refuses an unknown host and enrolls nothing | confirmed by running the real `ssh` binary, not inferred from the manpage |

If those ever drift, the two clients would maintain separate trust inside one file, each accepting keys the other had never seen, with nothing user-visible to show it. `TestOpenSSHRequiresGoTrustedKey` pins it by running the real openssh binary with the real option list against the test sshd: openssh must refuse first, the Go client enrolls, openssh must then accept that exact line, and neither may add a duplicate.

## verification

`go test ./cmd/project/` passes. The suite covers first-contact enrollment, an already-trusted key, a changed key for the same pod, per-pod separation, reload-before-enrollment, concurrent enrollment, a malformed store, failing closed mid-enrollment, newline repair on a hand-edited store, the 0700/0600 modes, the option list, and the openssh interop matrix.

Live run against a real CPU pod (`domra7onmjnwj9`, created and deleted, $0.06/hr):

| step | result |
| --- | --- |
| first `exec python` | `trusting new host key for pod … (SHA256:Mryb…)` on stderr, one entry written, `hello from the pod` on stdout |
| entry count after rsync ran | 1, so the openssh path used the Go client's entry instead of writing its own |
| second run | no notice, file byte-identical |
| tampered entry, Go client | exit 1, error names pod, fingerprint, path and remedy; store not healed |
| tampered entry, openssh directly | loud mismatch refusal, same fingerprint |
| genuine entry restored, openssh | connects, store still one line |

Legacy `exec` stdout is unchanged; every new message goes to stderr.

One scoping note on that table: it was run before the rsync path was strengthened from `accept-new` to `StrictHostKeyChecking=yes`, so the live rows exercised `accept-new`. The behaviour they demonstrate (matching pin accepted, mismatched pin refused) is identical under `yes`; the difference is only for an unknown host, and that case is covered against the real openssh binary by `TestOpenSSHRequiresGoTrustedKey`.

## residual risk, stated rather than papered over

Trust on first use prevents silent key substitution on later connections. It **cannot** prevent a machine-in-the-middle on the *first* connection to a pod. Genuine closure needs a host key, fingerprint, or ssh CA delivered through an authenticated Runpod control-plane channel, and that should be a separate ticket. CON-1158 should not be read as eliminating first-connect interception.

## scope

Deliberately a stopgap. [CON-874](https://linear.app/runpod/issue/CON-874) proposes deleting `cmd/project` outright, so this does not stand up a general trust store for a subsystem that may be removed.

## companion skill

The runpodctl skill's command reference gets a matching subsection, marked unreleased, in a separate PR against `runpod-plugins-official`. It documents that `ssh info` is unaffected and that a mismatched entry must never be deleted automatically.


donovanclarke (approved):
Approving. The trust-store design holds up: the alias/port-22 interop claim matches `knownhosts.New`'s unbracketed-pattern parsing and openssh's bare-alias lookup, the lock covers read/check/append and is never unlinked, every failure path fails closed without touching the store, and the interop test runs the real binary with the real option list. No critical or high findings. Three medium items below are worth a follow-up; two are about false "intercepted" alarms that undercut the PR's own rationale for pod-keyed trust.

**SEVERITY** - medium
**IMPACT** - Pod-keyed pins still false-alarm on every stop/start: the container disk resets and `runpod/containers` start.sh regenerates `/etc/ssh/ssh_host_*` at boot, so the same pod id presents a new key and the error calls it interception.
**SOLUTION** - Name stop/resume as the benign cause in the mismatch message and AGENTS.md/skill text, and reconsider whether pod-id keying actually delivers the stable identity the description claims.

**SEVERITY** - medium
**IMPACT** - Go negotiates `ecdsa-sha2-nistp256` first (`defaultHostKeyAlgos`), openssh prefers ed25519; interop only works because openssh reorders toward the recorded type. A user `HostKeyAlgorithms` setting, or a future x/crypto reorder, turns every existing pin into a "host key mismatch ... intercepted" error since `checkAddr` treats a different key type as a mismatch.
**SOLUTION** - When the alias already has an entry, set `ClientConfig.HostKeyAlgorithms` to that key's algorithm(s) (the skeema/knownhosts pattern) so the Go client pins the recorded type the way openssh does.

**SEVERITY** - medium
**IMPACT** - `Rsync`'s dry run captures ssh stderr into `dryRunBuf` but returns only `running rsync dry run: exit status N`, so an openssh refusal (unreadable store, key-type gap) surfaces without the fingerprint or `REMOTE HOST IDENTIFICATION HAS CHANGED` text this PR exists to show.
**SOLUTION** - Include `dryRunBuf` contents in the returned error (or stream stderr through) so the openssh diagnostic reaches the user.

<details open><summary><h1>Non-blocking (2)</h1></summary>

**SEVERITY** - low
**IMPACT** - rsync re-splits `-e` on spaces, and `StrictHostKeyChecking=yes` now makes a truncated `UserKnownHostsFile` fail closed, so a home directory with a space (common on Windows) breaks every rsync rather than degrading.
**SOLUTION** - rsync honours single/double quotes inside `-e`; wrap the `-i` and `UserKnownHostsFile` values in quotes in `getSshOptions`, or pass the options via a generated ssh config.

**SEVERITY** - low
**IMPACT** - `flock(LOCK_EX)` and `LockFileEx` without `FAIL_IMMEDIATELY` block indefinitely, so one wedged `project dev` process silently hangs every other `project`/`exec` invocation with no message.
**SOLUTION** - Acquire with `LOCK_NB`/`LOCKFILE_FAIL_IMMEDIATELY` in a short bounded retry loop and return a clear "host key store is locked by another runpodctl" error on timeout.

</details>

<details open>

review 1 · fable · [`5d0dfae`](https://github.com/runpod/runpodctl/commit/5d0dfaec6d2658940415e84af1acdbcf79b209df)

**RAN** - none (review only)

</details>

PR #344, opened 2026-09-15
fix(transfer): confine runpodctl receive to its destination (CON-1159)

OneAsterisk:
Closes [CON-1159](https://linear.app/runpod/issue/CON-1159). Last of the three sub-tickets under [CON-1158](https://linear.app/runpod/issue/CON-1158), alongside [#342](https://github.com/runpod/runpodctl/pull/342) (CON-1160) and [#343](https://github.com/runpod/runpodctl/pull/343) (CON-1161). All three touch disjoint files.

## the vulnerability

CWE-22. `receive` decoded a sender-controlled manifest and wrote wherever it said. The only defense was `strings.Contains(path, "..")`, which both over-rejects a file legitimately named `file..txt` and under-rejects absolute paths, Windows separators and symlinks.

**The ticket says this needs `--yes` or a known code. There is no `--yes` flag.** `cmd/transfer/transfer.go` hardcodes `NoPrompt: true, Overwrite: true`, so the receiver never sees the filenames before they are written. Any receiver holding the codephrase was exposed.

### measured, not argued

Same attack, both binaries, over the real relay. A symlinked subdirectory in the destination, then `send dir`:

| binary | exit | outcome |
| --- | --- | --- |
| `main` | **0** | silently wrote the payload **outside** the destination |
| this branch | 1 | refused, named the reason, wrote nothing |

That vector needs no hostile sender at all. The sender picks the path, so it picks which of the receiver's symlinks to write through. A symlinked download directory is enough.

## the fix, in two independent layers

Either layer alone would close the headline hole. Both exist because each covers something the other cannot, and that is the thing to preserve in review.

**`cmd/transfer/dest.go`** routes every receiver write through an `*os.Root` opened on the destination, so a path that leaves it fails at the syscall rather than relying on the caller having checked. The receive path now holds no path-based filesystem call at all:

```bash
grep -n 'os\.\(Stat\|Lstat\|MkdirAll\|OpenFile\|Create\|Symlink\|Remove\|ReadFile\|Open\)(' cmd/transfer/croc.go
```

The six remaining hits are all sender-side. Worth re-running on review.

**`cmd/transfer/confine.go`** validates the entire manifest before the first write. Archive entries go through the same `checkEntries`, so a zip cannot slip past whichever rule is laxer.

Two `os.Root` behaviors are why prevalidation is load-bearing rather than belt-and-braces. Both measured against the Go pin:

- `root.Symlink("../outside/x", "esc")` **succeeds.** Root then refuses to follow it, so nothing escapes during the transfer, but the receive has left an attacker-chosen link pointing out of the tree for whatever walks it next. `TestSymlinkCanEscapeTheRoot` asserts this gap deliberately, so it fails loudly if `os.Root` ever grows a guard.
- `root.Stat` on a pre-existing `dir -> /outside` returns `path escapes from parent`, **not** `ENOENT`. So every on-disk look is three-way, and the error arm never falls back to a plain `os.Stat`, which would reopen the hole the root was opened to close.

## also hardened, worth separate review attention

Outside the CWE-22 scope but in the same peer-input paths, so cheaper to fix now than to leave:

- `receiveData` had four panics on peer-controlled data, on a goroutine, so each took the process down rather than failing the transfer. `data[:8]` on a short frame was a slice bounds panic. The write offset is now bounded against the size the sender declared, which it was not, so a peer can no longer write outside the file it asked for or grow it without limit. A mutex leak on the write error path went with it.
- `utils.ChunkRangesToChunks` reads `chunkRanges[i+1]` while stepping by two from index one, so an even-length resume request is an out-of-range read, and it expands each count into that many entries, so a large one is an allocation the peer chose. `validateChunkRanges` rejects both. Checked against the upstream source.
- `TypeRecipientReady` asks a *sender* for a file. The receiver adopted the index it carried, which the loops after the transfer then used to index the manifest.

## behavior changes to weigh

Each was a silent success before, so none will show up as a test failure elsewhere.

1. **A symlinked subdirectory in the destination blocks a receive that writes into it.** This is the compat risk: symlinked download directories are common, and this is the change most likely to be reported as a bug. Following it is what the table above exploits.
2. **Sending a tree containing `link -> ../outside` is refused on arrival**, since the link would resolve outside the receive directory. In-tree symlinks transfer normally, and there is a test for that.
3. **An archive entry that is a symlink is refused** rather than written out as a regular file holding the target text, which is what upstream's extractor did.
4. **The empty-folder prompt no longer reads stdin under `NoPrompt`.** It did regardless, which was the only way a `receive` could hang.
5. **A refused archive is left in place** rather than deleted, so nothing that arrived is lost.

`send a.txt a.txt` still works. `GetFilesInfo` never deduped and the old receiver overwrote identical content harmlessly, so duplicates are allowed when kind, size, hash, target and the archive flag all agree, and refused otherwise. Verified over the live relay, not just in a unit test.

## why utils.UnzipDirectory could not be reused

It guards with `strings.Contains(path, "..")` after a `Clean`, so it rejects a legitimate `a..b` and misses a symlink-mode entry. It reads stdin to prompt about overwrites, which would hang a non-interactive receive. And it calls `log.Fatalln` on every error, taking the process down instead of returning. `extractZip` validates the whole archive before writing any of it, so a hostile archive cannot get half of itself onto disk.

## verification

`go test ./cmd/transfer/` passes: 76 tests, including six complete transfers over croc's own relay on loopback in about three seconds. `cmd/transfer` had no tests before this.

The hostile-sender tests build the manifest by hand, so `FolderRemote`, `Name`, `Symlink` and `TempFile` are all injectable. Two traps are documented in that file because the tests pass for the wrong reason without them:

- The sender hashes every entry before announcing the manifest, so naming a file that is not in `FolderSource` kills the sender, the receiver never sees the manifest, and "the escaping file does not exist" passes against completely unfixed code.
- **"Nothing escaped" is the weaker half of every assertion.** I removed the validator's `..` check to check the tests caught it: the traversal test still passed, because `os.Root` blocked the write and the transfer reported success having written nothing. Only the archive test failed. So each test now also asserts the receiver refused, which is what pins the validator rather than the root behind it. With both halves, removing that guard fails two tests.

Live e2e against the public relays: plain file, folder through the zip path with nested directories, `send a.txt a.txt`, receive into a non-empty directory, and the escaping-symlink case in the table above.

## one bug I introduced and caught

The fold-aware file-versus-directory check used exact keys at first, so a file `A` plus a directory `a/b` was accepted on APFS: one name doing both jobs. Found by exercising the validator against 43 cases before writing any tests. Case folding is measured with a filesystem probe rather than inferred from `GOOS`, cached per directory because ext4 carries the casefold attribute that way, and probed lazily because the probe of last resort has to write. A refused manifest leaves the destination untouched, and there is a test for that.

## cmd/croc

`cmd/croc/*.go` is deleted. Nothing imported it. It was a copy of the client retained when `cmd/transfer` was added, carrying every one of these bugs in parallel, which is the sort of thing that gets fixed once and then diverges silently.

`cmd/croc/relays.json` stays exactly where it is: `transfer.go` fetches it from that path on GitHub at runtime, so every released binary depends on the URL. There is now a comment there saying so, because a directory holding one JSON file and no Go looks like an oversight.

## companion skill

No command or flag surface changes, but `receive` gains refusals an agent could hit, and the symlinked-destination case is worth documenting. A follow-up to the skill repo should cover it once this lands.


donovanclarke (approved):
**Verdict: APPROVE.** Both layers are real and separately pinned: every receive-side filesystem call in `croc.go` goes through `confinedDest` (the remaining `os.*` calls in the head file are sender-side), and `validateManifest`/`validateArchive` run before the first write. I traced the manifest, declared-symlink, pre-existing-symlink, archive and `croc-stdin-` paths and could not construct a sender input that writes or links outside the destination. The `..`-through-declared-symlink lexical check, the three-way `os.Root` stat handling, the `refused` gate ahead of the peer's `TypeFinished`, and the receiver-side `TypeRecipientReady` guard all hold. The two medium items are about how a transfer fails inside the destination, not about confinement.

**SEVERITY** - medium
**IMPACT** - `receiveData` error paths (`decrypt`, short frame, offset bound, `WriteAt` failure such as disk full) now `return` from the data goroutine without signalling the control loop, so the transfer stalls until comm's read deadline instead of failing; before it at least exited.
**SOLUTION** - After logging, send `TypeError` on `c.conn[0]` and close it (or feed an error channel the `transfer()` loop selects on) so `Receive()` returns and both sides exit promptly.

**SEVERITY** - medium
**IMPACT** - Pre-existing, but the rewritten `recipientInitializeFile` keeps it: `recipientGetFileReady(finished=true)` still opens `filePaths[FilesToTransferCurrentNum]`, so when no file was transferred and entry 0 is a declared symlink, `openForWrite` follows the just-created link and `O_CREATE|O_TRUNC`/`Truncate(0)` hits its in-tree target; `send link` with `link -> data.bin` empties `dest/data.bin`.
**SOLUTION** - Return after the `TypeFinished` send when `finished` (the sender exits on `TypeFinished` and never reads the trailing `TypeRecipientReady`), or lstat the leaf and refuse to open through a symlink.

<details open><summary><h1>Non-blocking (7)</h1></summary>

**SEVERITY** - low
**IMPACT** - Pass four checks ancestors only and `updateIfRecipientHasFileInfo` uses `lstat`, so a pre-existing escaping symlink at the leaf (`dest/a.txt -> /etc/x`) survives validation; `openForWrite` then errors inside `recipientGetFileReady`, whose result is discarded at the `//nolint` call, so `TypeRecipientReady` never goes out and both peers hang instead of the table's promised refusal.
**SOLUTION** - Lstat the leaf of every `kindFile` entry in pass four and refuse a symlink there, and propagate `recipientGetFileReady`'s error.

**SEVERITY** - low
**IMPACT** - `writeZipEntry` honours `entry.Mode().Perm()` from the archive while `dest.go` deliberately ignores the peer-controlled `FileInfo.Mode`; a hostile zip can create `0o777` files (umask permitting), which is inconsistent with the stated rule.
**SOLUTION** - Use `receivedFilePerm`, or derive only the executable bit from the entry and mask the rest.

**SEVERITY** - low
**IMPACT** - `positionInt64+int64(len(payload)) > declaredSize` wraps negative for offsets near `MaxInt64` and passes the bound; the OS then rejects the `pwrite`, so nothing escapes, but the check is not what it claims.
**SOLUTION** - Write it as `positionInt64 > declaredSize-int64(len(payload))` after the `len(payload) <= declaredSize` check.

**SEVERITY** - low
**IMPACT** - Pass four only inspects ancestors, so a declared file whose `rel` is an existing directory fails at `openForWrite` mid-transfer after earlier files landed, and a declared empty folder whose `rel` is an existing file is silently skipped (`isEmptyDir` error falls into `continue`), both breaking all-or-nothing for user-side state.
**SOLUTION** - Lstat each leaf in pass four and refuse a file-over-directory or directory-over-file before the first write.

**SEVERITY** - low
**IMPACT** - The fold probe measures case only; APFS also aliases Unicode normalization forms (NFC/NFD) and Windows aliases trailing dots/spaces and 8.3 short names, so "same destination, different contents" can be bypassed there and the last write wins, in-tree only.
**SOLUTION** - Note the gap beside the probe, or key `lowered` on a normalized form as well as lowercase.

**SEVERITY** - low
**IMPACT** - `maxResumeChunks` at `1<<24` still lets a receiver make the sender allocate a ~16M-entry `chunkMap` plus a 128MB `CurrentFileChunks` slice, hundreds of MB chosen by the peer.
**SOLUTION** - Bound `total` by `ceil(FilesToTransfer[n].Size/ranges[0])`, which is known once the file index is validated.

**SEVERITY** - low
**IMPACT** - `startTestRelay` calls `t.Fatalf` inside `sync.Once`; if the relay fails to come up during the first test, every later test indexes an empty `relayPorts` and panics with an index error rather than reporting the cause.
**SOLUTION** - Record the startup error in a package var and `t.Fatalf` it from each caller after `Once.Do`.

</details>

<details open>

review 1 · fable · [`ab63fb5`](https://github.com/runpod/runpodctl/commit/ab63fb5516c2a3a763a6ed36d52edd357601367a)

**RAN** - none (review only)

</details>

